PR #1, opened 2023-11-23
add for rh-demo

sallyom:
- added a new Containerfile with rh-python
- added a var for MODEL_FILE to customize which one to use

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -5,5 +5,6 @@ COPY requirements.txt /locallm/requirements.txt
 RUN pip install --upgrade pip
 RUN pip install --no-cache-dir --upgrade -r /locallm/requirements.txt
 USER user
+ENV MODEL_FILE=llama-2-7b-chat.Q5_K_S.gguf
 COPY --chown=1000 ./ /locallm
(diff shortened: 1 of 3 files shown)


chat-rh.py
```diff
@@ -0,0 +1,34 @@
...
+class Chat:
+
+    n_ctx = 2048
+
+    def __init__(self) -> None:
+        self.chat_history = [
+                {"role": "system", "content": """You are a helpful assistant that is comfortable speaking
+                 with C level executives in a professional setting."""},
+                ]
+        self.llm = Llama(model_path="/locallm/nested/dir/llama-2-7b-chat.Q4_K_M.gguf",
```
sallyom: that #L12 is the only line that is different from chat.py

MichaelClifford: Might be easier to just add `model_path` parameter to the the chat class `__init__`, instead of managing another file.

sallyom: yup was thinking that but also was being lazy

MichaelClifford (approved):
LGTM! Thanks for the PR @sallyom 🎉

PR #6, opened 2023-12-14
update containerfile with download

sallyom:
@MichaelClifford ok to keep this Containerfile in the repo? I use it so it's convenient :)


sallyom:
I added 2 files in `loadmodel` dir, 
1) just the model in a micro image <- we can add an apiserver & it can build with any provided model
2) model + app

MichaelClifford (approved):

PR #7, opened 2023-12-21
add summarize app

MichaelClifford:
Updates the `Chat` class to include a `summarize()` method. Also includes the a new `summary_app.py` and `Containerfile` along with a README for instructions on how to build and use it.


MichaelClifford:
This PR changed significantly since the first commit with the intention of making the summarizer app more independent of other components of the repo. This resulted in a new directory `summarizer` that contains the python scripts for the model service, an example ai application that  uses the service and the necessary artifacts to build an image.

PR #10, opened 2024-01-05
fix bug in summarizer imports

MichaelClifford:


summarizer/model_services/summary_service.py
```diff
@@ -2,9 +2,9 @@
 sys.path.append("src")
 import gradio as gr
 from llama_cpp import Llama
-from chat import clip_history, chunk_tokens
+from llamacpp_utils import clip_history, chunk_tokens
  
-llm = Llama("llama-2-7b-chat.Q5_K_S.gguf",
+llm = Llama("models/llama-2-7b-chat.Q5_K_S.gguf",
```
jeffmaury, line 7: Change does not seem required to me as the image does not start anymore because the model cannot be found

MichaelClifford:
https://github.com/MichaelClifford/locallm/tree/main/summarizer#download-models

The code assumes the models are stored in the `/models` directory.

MichaelClifford:
nvm, I see why this is not working... Containerfile needs to be updated as well. Sorry. PR incoming.

PR #15, opened 2024-01-10
Updates to READMEs & Containerfiles

sallyom:
this PR:

* removes chatbot,model_services/builds/arm,x86
* moves Containerfile to ./model_services/base/Containerfile
* adds a chatbot/model_services/cuda
* moves the current x86/Containerfile under chatbot/model_services/cuda/ 
* updates READMEs
* adds docs for how to run sample app alongside model-service
* adds `gitrepo` to ai-studio.yaml - so all that's required to pass to podman-desktop is the single `ai-studio.yaml` file, then, `contextdir` is relative to the root of the gitrepo, and `containerfile` is relative to `contextdir`
* adds `backend` to ai-config to distinguish between multiple images with same arch
* adds `gpu-env` to distinguish images by gpu environments

note:
* We talked about stripping the 'sample-app' logic from the model-service Containerfiles - that can be a follow-up PR
* I built the images and pushed to `quay.io/sallyom/model-services`, see the `ai-studio.yaml` where i have those images listed/commented.


chatbot/ai-studio.yaml
```diff
@@ -1,16 +1,31 @@
...
       arch: amd64
       model-service: true 
+      #image-name: quay.io/sallyom/model-services:llama2-7b-chatbot
+    - name: chatbot-model-x86-cuda
+      contextdir: model-services/builds/cuda
+      arch: amd64
+      model-service: true 
+      #image-name: quay.io/sallyom/model-services:llama2-7b-chatbot-cuda
+    - name: chatbot-example-arm
+      contextdir: ai_applications/builds
```
sallyom: for our sample apps, we don't need these `ai_applications/builds/Containerfile` images, the model-service images alone currently provide what is necessary for a simple sample app.

sallyom: I'm not suggesting we remove them, we'll document how to build/run these, but they don't need to be included in what we build as the sample app in `ai-studio.yaml`

sallyom: or, on 2nd thought, we can keep these in, because it will provide the "this is how you develop an application to interact with the model-service"

summarizer/model_service/builds/base/Containerfile
```diff
@@ -0,0 +1,14 @@
+# Pass build args to build a specified model from a specified local directory
+FROM registry.access.redhat.com/ubi9/python-39:latest
+WORKDIR /locallm
+ARG MODEL_FILE=llama-2-7b-chat.Q5_K_S.gguf
+ARG MODEL_LOCAL_DIR=models
+COPY $MODEL_LOCAL_DIR/$MODEL_FILE /locallm/models/
```
sallyom: @MichaelClifford will these ARGS work? I thought we needed to export just the MODEL_FILE env var so kept the dir/file separate args - but maybe that's not required.

chatbot/ai-studio.yaml
```diff
@@ -1,16 +1,30 @@
...
+  name: chatbot
+  description: This is a LLM chatbot application
   containers:
-    - name: chatbot-example
-      contextdir: ai-application/builds
-      #image-name: quay.io/redhat-et/chatbot:latest
+    # podman run -p 7860:7860 quay.io/sallyom/model-services:llama2-7b-chatbot-arm
     - name: chatbot-model-arm
-      contextdir: model-services/builds/arm
+      contextdir: chatbot/model_service/builds/base
```
sallyom: I'm not sure how contextdir is used with desktop - will there be a 
`cd $contextdir` then
`podman build .`
or
does `contextdir` only mean "this is where the Containerfile is" and all builds happen from the root of the repository? 
ie - podman-desktop is supplied with the `url path to this ai-studio.yaml` _and_ the `git repository URL`? If so, is contextdir required here?

chatbot/ai-studio.yaml
```diff
@@ -1,16 +1,30 @@
...
+      #image-name: quay.io/sallyom/model-services:llama2-7b-chatbot-cuda
+    - name: chatbot-example-arm
+      contextdir: chatbot/ai_applications/builds
+      arch: arm64
+      #image-name: quay.io/sallyom/chatbot:latest-arm
+    - name: chatbot-example-x86
+      contextdir: chatbot/ai_applications/builds
+      arch: amd64
+      #image-name: quay.io/sallyom/chatbot:latest
+
```
sallyom: @MichaelClifford I used the Containerfiles from this PR to build the images above, so I'm confident they build (we'll get some GH actions going asap that will sanity-check the image builds w/ each PR). 
These are good as/of right now from this PR:
- chatbot/model_service/builds/base/Containerfile  --platform linux/amd64 && --platform linux/arm64
- summarizer/model_service/builds/base/Containerfile  --platform linux/amd64 && --platform linux/arm64
- chatbot/model_service/builds/cuda/Containerfile --platform linux/amd64
- summarizer/model_service/builds/cuda/Containerfile --platform linux/amd64
Feel free to pull the arm images down and try them.

chatbot/ai-studio.yaml
sallyom: In [#17](https://github.com/redhat-et/locallm/pull/17) there's a `containerfile` field added - so here the values would be:
```
contextdir: chatbot/model_services
containerfile: base/Containerfile
```

lstocchi:
@sallyom @MichaelClifford adding it here but it can be handled discusses somewhere else. Another thing that would be useful for us would be to add a version value to the ai-studio.yaml file. So if in future there is some breaking changes we can check the version and know how to parse the file/which schema to use.

chatbot/ai-studio.yaml
```diff
@@ -1,16 +1,36 @@
...
-      contextdir: model-services/builds/x86
+      contextdir: .
+      containerfile: chatbot/model_service/builds/base/Containerfile
+      arch: amd64
+      model-service: true 
+      #image-name: quay.io/sallyom/model-services:llama2-7b-chatbot
+    - name: chatbot-model-x86-cuda
+      contextdir: .
+      containerfile: chatbot/model_service/builds/cuda/Containerfile
       arch: amd64
```
lstocchi: What we do now is parsing the file and get all the values with the `arch` that fits the user machine. If there are two entries of the same app that uses the same arch it would be hard to guess which to choose. We would need another info to specify the hardware accelerator, if any. Either an additional property or a different name like amd64-cuda.

summarizer/model_service/builds/base/Containerfile
sallyom: never mind that ^^ now that we aren't copying the models into the image this is N/A

chatbot/ai-studio.yaml
sallyom: This brings up the need for more thought around this configuration file - how about a new field like `backend` so like 
```
- name: chatbot-model-x86-cuda
  contextdir: .
  containerfile: chatbot/model_service/builds/x86-cuda/Containerfile
  arch: amd64
  backend: cuda (or llama, stablediffusion, etc,etc)
  ```
 see: https://github.com/go-skynet/model-gallery/tree/main for different examples.)  I'll add `backend` for now.

File: chatbot/ai-studio.yaml (hunk 3 above)
Comment (sallyom) [reply chatbot/ai-studio.yaml]: see note below also, about adding `gitrepo` and `backend` fields

File: chatbot/ai-studio.yaml (hunk 5 above)
Comment (sallyom) [reply chatbot/ai-studio.yaml]: the `backend` will only be relevant for those images where `model-service: true` I suppose

Comment (sallyom) [reply chatbot/ai-studio.yaml]: ie, you could filter for `model-service` -> `backend`

**Comment by sallyom** [comment]
@MichaelClifford I don't think we're using the `Makefile` - we could update it but rn it doesn't match our docs, should we remove it for now?

File: chatbot/ai-studio.yaml (hunk 5 above)
Comment (MichaelClifford) [reply chatbot/ai-studio.yaml]: I don't think "cuda" and "llama" are in the same category. Right now the only "backend" (i've been using the term framework) we want to support is "llama" (which I assume is short for llama.cpp). "cuda" would be something like "accelerator" or "gpu_type" I think. Right now the only that we support (I think) are cuda and eventually Apple Metal.

Comment (MichaelClifford) [reply chatbot/ai-studio.yaml]: Also, here "arm" and "x86" should just be collapsed into one thing right? Since they use the same container file and everything. This will just be the basic image and for now the only special image will be the cuda-x86.

File: chatbot/model_service/builds/x86-cuda/Containerfile [hunk 6]
```diff
@@ -1,10 +1,14 @@
 FROM quay.io/opendatahub/workbench-images:cuda-ubi9-python-3.9-20231206
 WORKDIR /locallm
-COPY builds/requirements.txt /locallm/requirements.txt
+# MODEL_FILE should be set at runtime
+ENV MODEL_FILE=llama-2-7b-chat.Q5_K_S.gguf
```
MichaelClifford: Is this used for anything any more? I think "MODEL_PATH" might be sufficient.

chatbot/model_service/builds/x86-cuda/Containerfile
```diff
@@ -1,10 +1,14 @@
...
+ENV MODEL_PATH=/locallm/models/${MODEL_FILE}
+COPY chatbot/model_service/builds/requirements.txt .
```
MichaelClifford: I'm not convinced that we can run `podman build` from the top directory without a major lag due to copying the entire repo, specifically the models dir. Is there any other way around this? Maybe push all other code down one level and the top dir is just like "src" and "models" the we build everything from "src" dir?

chatbot/model_service/chat_service.py
```diff
@@ -7,7 +7,7 @@
  
 
 llm = Llama(os.getenv('MODEL_PATH', 
-                      "models/llama-2-7b-chat.Q5_K_S.gguf"),
+                      "/locallm/models/llama-2-7b-chat.Q5_K_S.gguf"),
```
MichaelClifford: isn't the working dir already `locallm`? Is this necessary?

chatbot/ai-studio.yaml
lstocchi: @MichaelClifford if a model service app supports llama, it would work with any llama models? If so, i think we could end up having different model services based on the "backend" used like 
```
model_service1:
   backend: 
     - llama
     - b
model_service2:
   backend:
     - c
     - ....
```
 and, consequently, i agree with using the term "gpu_type" or "accelerator" for cuda or apple metal.

chatbot/model_service/builds/base/Containerfile
```diff
@@ -0,0 +1,12 @@
+FROM registry.access.redhat.com/ubi9/python-39:latest
+# MODEL_FILE should be set at runtime
+ENV MODEL_FILE=llama-2-7b-chat.Q5_K_S.gguf
+ENV MODEL_PATH=/locallm/models/${MODEL_FILE}
```
lstocchi: Are these really needed here? We are mounting volume at runtime so we have to set them with the correct path/name 
Maybe we should add this kind of info somewhere in the ai-studio or agree that every model service from now on would need the env MODEL_PATH and the sample application MODEL_ENDPOINT, so we can set them at runtime. And we need to document it

chatbot/model_service/chat_service.py
sallyom: yes, workdir is /locallm

chatbot/ai-studio.yaml
MichaelClifford: yeah so our current model service uses llama.cpp as the backend, which supports  any of the models listed [here](https://github.com/ggerganov/llama.cpp?tab=readme-ov-file#description).

If we wanted a different backend, I think that would require different image. So yes,I think we'd have different model services based on different backends. 

We probably need to build out a second "backend" to see how this would actually work though.

MichaelClifford:
> @MichaelClifford I don't think we're using the `Makefile` - we could update it but rn it doesn't match our docs, should we remove it for now?

yeah, that was something I was using a while back. We can probably remove it. But that can also happen in a separate PR.

chatbot/model_service/builds/base/Containerfile
MichaelClifford: yes, I don't think we need these any longer as it will be required to define model_path at runtime instead.

chatbot/ai-studio.yaml
sallyom: should we add both fields? 
```
backend: llama
gpu-env: cuda
```

chatbot/model_service/builds/x86-cuda/Containerfile
sallyom: we can do that, I'll move things around

chatbot/model_service/builds/base/Containerfile
sallyom: MODEL_PATH is used in the python code, yea - I figured I'd set them here as a default but it's probably better to remove them, so I will do that

chatbot/model_service/builds/x86-cuda/Containerfile
sallyom: I removed both, they aren't needed for the builds

chatbot/model_service/builds/x86-cuda/Containerfile
sallyom: I updated - only the chatbot & we can update the summarizer in a followup - 
```
chatbot/model_services/base
chatbot/model_services/cuda
chatbot/model_services/src
```
and the shared files live in src (requirements.txt, chat_service.py, llamacpp_utils.py)

chatbot/model_service/chat_service.py
sallyom: removed it

chatbot/model_service/builds/base/Containerfile
sallyom: removed these

sallyom:
@MichaelClifford I made this PR smaller - split up other changes into follow-ups - went from 24 files changed to 12, ptal!

lstocchi:
@sallyom @MichaelClifford if you agreed on building from the root and having separate entries for the same Containerfile for different arch, it would be good to have this merged so we can iterate on this. I really feel sorry to push new PRs and make Sally rebase it again :)

Thanks @sallyom for the patience and the hard work 🙏 🎉

MichaelClifford (approved):
LGTM

MichaelClifford:
Thanks @sallyom 😄

PR #17, opened 2024-01-16
fix: update chatbot ai-studio.yaml and containerfile

lstocchi:
This PR updates the Chatbot app by adding/updating 4 things:

1) it makes the old arm containerfile the default one and makes the old x86 the containerfile for amd64-cuda hardware
2) it updates the ai-studio to represent the new structure and adds a new property `containerfile` so to know where it is within the folder and its actual name
3) it adds the `llamacpp_utils.py` inside the builds folder so that we do not need to run the `build.sh` script from podman-desktop after cloning the repo (this part can be removed if you prefer not to have duplicate files)
4) update the build.sh script and doc

### DIFF
--- a/chatbot/model_services/builds/build.sh
+++ b/chatbot/model_services/builds/build.sh
@@ -1,12 +1,6 @@
 #!/usr/bin/ bash
 
-MODEL=${1:-llama-2-7b-chat.Q5_K_S.gguf}
-ARCH=${2:-arm}
-IMAGE_NAME=${3:-locallm}
+IMAGE_NAME=${1:-locallm}
+ARCH=${2:-.}
 
-echo "building with $MODEL on $ARCH"
-cp ../../../models/$MODEL .
-cp -r ../../../src .
 podman build -t $IMAGE_NAME .. -f $ARCH/Containerfile
-rm $MODEL
-rm -rf src
--- a/chatbot/model_services/builds/src/llamacpp_utils.py
+++ b/chatbot/model_services/builds/src/llamacpp_utils.py
@@ -1,0 +1,31 @@
+from llama_cpp import Llama
+
+
+def tokenize(llama, prompt):
+    return llama.tokenize(bytes(prompt, "utf-8"))
+
+def count_tokens(llama,prompt):
+    return len(tokenize(llama,prompt)) + 5
+
+def clip_history(llama, prompt, history, n_ctx, max_tokens):
+    prompt_len = count_tokens(llama, prompt)
+    history_len = sum([count_tokens(llama, x["content"]) for x in history])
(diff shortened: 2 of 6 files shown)


chatbot/model_services/builds/src/llamacpp_utils.py
```diff
@@ -0,0 +1,31 @@
+from llama_cpp import Llama
```
MichaelClifford, line 1: I'm not sure we want to have a copy of this in every model service directory. Is there another way to manage it so this can be a share file among all the model services in the repo?

chatbot/model_services/builds/build.sh
```diff
@@ -1,12 +1,6 @@
 #!/usr/bin/ bash
```
MichaelClifford, line 2: If we cut it down to this, its probably not needed anymore. We can just provide the direct `podman build` instructions in the docs.

chatbot/model_services/builds/src/llamacpp_utils.py
MichaelClifford: Can we make the top directory the build context and have podman ignore the `models/` so that it does not bring in any of those larger files? That way we can still share the utils files among the different model services?

sallyom: ~we can use a symlink from the root of the repo src~ see note below

sallyom:
@lstocchi @MichaelClifford [#15](https://github.com/redhat-et/locallm/pull/15) has updates that are relevant to this PR, it removes the need for the `build.sh` and it assumes you start the build from the root of this repository rather than from chatbot/model-service/builds directory - so the `src/llamacpp_utils.py` is shared for all builds (as well as the models dir, but that is configurable as a build-arg)

I think podman-desktop should only expect the url to the single `ai-studio.yaml` file - with an arbitrary name - and should include a `gitrepo` - so this file could live outside of the associated git repository  - this sets up the ability for a central index that holds a library of these files - similar to how [krew-index](https://github.com/kubernetes-sigs/krew-index/tree/master/plugins) is organized to reference git repos that hold kubectl plugins

(I added the `containerfile` and `gitrepo` fields in the ai-studio.yaml in #15)

MichaelClifford:
@sallyom if we assume the root directory as the build context, won't podman desktop copy the entire repo into the vm before starting the build? In which case all models stored in "models" will be pulled in, which could be quite slow.

MichaelClifford (approved):

MichaelClifford:
I'm going to go ahead an merge this one now to unblock @lstocchi and team, but there may be some additional upcoming changes from #15 incorporated too.

PR #18, opened 2024-01-18
add playground build

MichaelClifford:
Adds a a new directory "playground" that contains the `requirements.txt`,`run.sh` and `Containerfile` needed to build a simple llama.cpp compatible model server image. 

The image assumes that the model will be volume mounted at runtime, and so does not include any models. The image is only about ~1Gb.

### DIFF
--- a/playground/Containerfile
+++ b/playground/Containerfile
@@ -1,0 +1,7 @@
+FROM registry.access.redhat.com/ubi9/python-39:latest
+WORKDIR /locallm
+COPY requirements.txt /locallm/requirements.txt
+RUN pip install --upgrade pip
+RUN pip install --no-cache-dir --upgrade -r /locallm/requirements.txt
+COPY run.sh run.sh
+ENTRYPOINT [ "sh", "run.sh" ]
--- a/playground/requirements.txt
+++ b/playground/requirements.txt
@@ -1,0 +1,1 @@
+llama-cpp-python[server]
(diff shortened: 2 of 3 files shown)


playground/Containerfile
```diff
@@ -0,0 +1,7 @@
+FROM registry.access.redhat.com/ubi9/python-39:1-158
```
sallyom: `registry.access.redhat.com/ubi9/python-39:latest` since registry.access.redhat.com keeps their latest tags up-to-date - so we won't have to update it here
for example, that `1-158` tag is already not the most recent, I checked:
1-158: "Created": "2023-11-27T14:09:22.968117784Z"
latest: "Created": "2023-12-07T06:33:01.387970424Z"

sallyom (commented):
with `registry.access.redhat.com` I'd switch to the `python-39:latest` tag instead of the pinned tag (or if pinning to a specific, we should use the @sha256 digest instead of tags - but then we have the burden of keeping it updated)

lstocchi:
how can we interact with it? Do we need to send a post request? Is it like local-ai that accepts a folder with multiple volumes and we need to tell which one to use?

MichaelClifford:
You should be able to run it with `podman run -it -p 8000:8000 -v <Local Path>/locallm/models:/locallm/models:Z -e MODEL_PATH=models/llama-2-7b-chat.Q5_K_S.gguf playground`

and then interact with it with:

```
curl http://localhost:8000/v1/completions -H "Content-Type: application/json" -d '{
     "model": "your-model.gguf",                    
     "prompt": "A long time ago in a galaxy far, far away",
     "temperature": 0.7
   }'
```

lstocchi (approved):
LGTM

PR #20, opened 2024-01-23
fix: merge chatbot config files into one

lstocchi:
This PR just revert the change made and create a single ai-studio file.


chatbot/ai-studio.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - name: chatbot-model-service
+      contextdir: model_services
+      containerfile: base/Containerfile
+      model-service: true
+      backend: 
+        - llama
+      arch:
+        - arm64
+        - amd64
+    - name: chatbot-model-servicecuda
```
jeffmaury, line 18: seems to me there is an issue as if I run on amd64 and cuda I will get 2 containers being elected or we should revert the order and consider the name of the container is the same to filter

sallyom: filtering by name of container won't scale - users will be providing podman-desktop their own ai yamls, filtering by arch seems better to me

sallyom: filter by arch && gpu?

chatbot/ai-studio.yaml
```diff
@@ -0,0 +1,27 @@
+application:
+  type: language
+  name: chatbot
+  description: This is a LLM chatbot application that can interact with a llamacpp model-service
+  containers:
+    - name: chatbot-inference-app
+      contextdir: ai_applications
+      containerfile: builds/Containerfile
```
sallyom, line 8: the chatbot-inference-app also should include arm64 && amd64 arch

chatbot/ai-studio.yaml
lstocchi: my idea was that we have to select the one closer to the user' system. If they have cuda and amd64 we'll pick cuda otherwise the default amd64.
If there is one that support my gpu i don't see why we need to pick the one that don't support it

chatbot/ai-studio.yaml
lstocchi: done 👍

chatbot/ai-studio.yaml
```diff
@@ -6,6 +6,9 @@ application:
     - name: chatbot-inference-app
       contextdir: ai_applications
       containerfile: builds/Containerfile
+      arch:
```
jeffmaury: I don't think we need an arch field here

lstocchi: I don't have a strong opinion on it, so i added it on Sally's suggestion but ...can there be a case where the sample app has two version, one for amd and one for arm? In that case it may be useful. Or we could state that if there is no arch field is good for any arch.

jeffmaury: Yes my comment is just as the client application is a full Python app it has no arch adherence

lstocchi: @jeffmaury removed the last commit having the update.

chatbot/ai-studio.yaml
lstocchi: closing as i think it is resolved after the discussion we had with Jeff. 
We're going to filter with arch && gpu

jeffmaury (approved):
LGTM

PR #22, opened 2024-01-24
add scratch model image containerfile

sallyom:
This adds a Containerfile for packaging/containerizing a model (for using in a disk image) 
/cc @lstocchi 
/cc @MichaelClifford


MichaelClifford:
This only adds the model URL right, not any actual model?

sallyom:
@MichaelClifford it downloads the model! I think so at least, it's 4.65G

MichaelClifford (approved):

PR #23, opened 2024-01-25
update summarizer dir for podman AI studio

MichaelClifford:
Updates the `summarizer` directory to conform with the rest of the recipes in the repo


summarizer/ai_applications/base/Containerfile
```diff
@@ -0,0 +1,7 @@
+FROM registry.access.redhat.com/ubi9/python-39:1-158
```
sallyom, line 1: make that python-39:latest (can do in a follow-up if this is the only nit)

sallyom (approved):
LGTM! I'll update that image tag in a follow-up, will be adding some docs tomorrow

PR #24, opened 2024-01-25
update summarizer containerfiles

MichaelClifford:


summarizer/ai_applications/base/Containerfile
```diff
@@ -1,7 +1,7 @@
 FROM registry.access.redhat.com/ubi9/python-39:1-158
-WORKDIR /locallm
-COPY requirements.txt /locallm/requirements.txt
+WORKDIR /summarizer
```
sallyom, line 2: change to `/locallm` (or update #L5, but probably better to keep them all the same for now)

sallyom: actually, I see what you did, since this isn't actually a locallm image, you made the WORKDIR more generic :)

summarizer/ai_applications/base/Containerfile
```diff
@@ -1,7 +1,7 @@
...
+COPY base/requirements.txt .
 RUN pip install --upgrade pip
 RUN pip install --no-cache-dir --upgrade -r /locallm/requirements.txt
```
sallyom: change to /summarizer/requirements.txt

sallyom (commented):
1 thing, then lgtm

summarizer/ai_applications/base/Containerfile
MichaelClifford: the `requirements.txt` is in `summarizer/base/requirement's.txt` the same way as in the other directories. Should it be in a different location?

lstocchi:
same as Sally. Just an update on L5 `/locallm/` =>  `/summarizer/` https://github.com/redhat-et/locallm/pull/24/files#diff-ee4a5788fe9206a9b7f41b4f51d598b69729a0ad53c8475ef50158d0da9cdab0L5 and it works fine 👍

summarizer/ai_applications/base/Containerfile
MichaelClifford: nvm. I see where the error is.

MichaelClifford:
Thanks for the reviews @sallyom @lstocchi . Should be good to go now.

lstocchi (approved):
LGTM

PR #29, opened 2024-01-26
Add chatbot quadlet

sallyom:
This adds a quadlet file and a pod yaml for the chatbot application to run it as a systemd service

chatbot/quadlet/README.md gives the commands to run it with podman.


models/Containerfile
```diff
@@ -1,7 +1,7 @@
 #https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGUF/resolve/main/llama-2-7b-chat.Q5_K_S.gguf
-#https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/blob/main/mistral-7b-instruct-v0.1.Q4_K_S.gguf
+#https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_S.gguf
 # podman build --build-arg MODEL_URL=https://... -t quay.io/yourimage .
-FROM scratch
+FROM registry.access.redhat.com/ubi9/ubi-micro:9.3-13
```
sallyom, line 4: I had to use the ubi-micro - because a shell is required in the initContainer in order to share the model-file with the model-service container.

chatbot/quadlet/chatbot.yaml
```diff
@@ -0,0 +1,40 @@
...
+kind: Pod
+metadata:
+  labels:
+    app: chatbot-inference-app
+  name: chatbot-inference-app
+spec:
+  initContainers:
+  - name: model-file
+    image: quay.io/sallyom/models:llama2-7b-gguf
+    command: ['/usr/bin/bash', '-c', "cp /model/llama-2-7b-chat.Q5_K_S.gguf /shared/ && chmod 777 /shared/llama-2-7b-chat.Q5_K_S.gguf"]
```
rhatdan, line 11: Don't know what this file is, or why it needs to be chowned 777 also why is it being copied?
It would have been better to do this in one command with 
install -m 777 ...

But I still question what is going on here?

rhatdan: @ygalblum PTAL, we are trying to mount an image into a volume, is there a preferred way to do this in k8s?

ygalblum: @rhatdan we were having a similar discussion in this this issue https://github.com/containers/podman/issues/16819. AFAIK, and from what I could gather, there is no official way to do it. As a result, the recommended way is the one implemented here. I.e. use an init-container to copy data into a shared volume.

Having said that, if the intention is to use this Yaml file only with Podman, Podman does support creating a volume out of an image. Unfortunately, the current version does not support this feature in a K8S Yaml, but https://github.com/containers/podman/pull/21342 will be includes in the next one.

PR #31, opened 2024-01-30
RAG, Langchain and Chromadb example 

MichaelClifford:
This PR adds a new recipe `rag-langchain` to the repo. In this example we build only the ai-application code and rely on the existing playground image for our model service and the external chormaDB image for our vectorDB service.

### DIFF
--- a/rag-langchain/ai-studio.yaml
+++ b/rag-langchain/ai-studio.yaml
@@ -15,6 +15,7 @@ application:
     - name: chromadb-server
       contextdir:: builds/chromadb
       containerfile: Containerfile
+      vectordb: true
       arch:
         - arm64
         - amd64
--- a/rag-langchain/ai-studio.yaml
+++ b/rag-langchain/ai-studio.yaml
@@ -1,0 +1,22 @@
+application:
+  type: language
+  name: rag-demo
+  description: This is a RAG demo application. 
+  containers:
+    - name: llamacpp-server
+      contextdir: playground
+      containerfile: Containerfile
+      model-service: true
+      backend: 
+        - llama
+      arch:
+        - arm64
+        - amd64
+    - name: chromadb-server
+      image: docker.io/chromadb/chroma
+    - name: rag-inference-app
+      contextdir: rag-langchain
+      containerfile: builds/Containerfile
+      arch:
+        - arm64
+        - amd64
--- a/rag-langchain/ai-studio.yaml
+++ b/rag-langchain/ai-studio.yaml
@@ -4,7 +4,7 @@ application:
   description: This is a RAG demo application. 
   containers:
     - name: llamacpp-server
-      contextdir: playground
+      contextdir: ../playground
       containerfile: Containerfile
       model-service: true
       backend: 
@@ -15,7 +15,7 @@ application:
     - name: chromadb-server
       image: docker.io/chromadb/chroma
     - name: rag-inference-app
-      contextdir: rag-langchain
+      contextdir: .
       containerfile: builds/Containerfile
       arch:
         - arm64
--- a/rag-langchain/ai-studio.yaml
+++ b/rag-langchain/ai-studio.yaml
@@ -13,7 +13,11 @@ application:
         - arm64
         - amd64
     - name: chromadb-server
-      image: docker.io/chromadb/chroma
+      contextdir:: builds/chromadb
+      containerfile: Containerfile
+      arch:
+        - arm64
+        - amd64
     - name: rag-inference-app
       contextdir: .
       containerfile: builds/Containerfile
--- a/rag-langchain/rag_app.py
+++ b/rag-langchain/rag_app.py
@@ -1,0 +1,89 @@
+from langchain_openai import ChatOpenAI
+from langchain_core.prompts import ChatPromptTemplate
+from langchain_core.runnables import RunnablePassthrough
+from langchain.embeddings.sentence_transformer import SentenceTransformerEmbeddings
+from langchain.callbacks import StreamingStdOutCallbackHandler
+from langchain.text_splitter import CharacterTextSplitter
+from langchain_community.document_loaders import TextLoader
+from langchain_community.vectorstores import Chroma
+
+from chromadb import HttpClient
+from chromadb.config import Settings
+import chromadb.utils.embedding_functions as embedding_functions
+
+
+import uuid
+import os
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-d", "--docs", default="data/fake_meeting.txt")
+parser.add_argument("-c", "--chunk_size", default=150)
+parser.add_argument("-e", "--embedding_model", default="BAAI/bge-base-en-v1.5")
+parser.add_argument("-H", "--vdb_host", default="0.0.0.0")
+parser.add_argument("-p", "--vdb_port", default="8000")
+parser.add_argument("-n", "--name", default="test_collection")
+parser.add_argument("-m", "--model_url", default="http://0.0.0.0:8001/v1")
+
+args = parser.parse_args()
+llm = ChatOpenAI(base_url=args.model_url, 
+                 api_key="EMPTY",
+                 streaming=True,
+                 callbacks=[StreamingStdOutCallbackHandler()])
+
+prompt = ChatPromptTemplate.from_template("""Answer the question based only on the following context:
+{context}
+
+Question: {input}
+"""
+)
+
+### populate the DB ####
+
+#os.environ["HF_HUB_CACHE"] = "./models/"
+os.environ["TOKENIZERS_PARALLELISM"] = "false"
+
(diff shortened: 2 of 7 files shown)


rag-langchain/ai-studio.yaml
```diff
@@ -0,0 +1,22 @@
+application:
+  type: language
+  name: rag-demo
+  description: This is a RAG demo application. 
+  containers:
+    - name: llamacpp-server
+      contextdir: playground
```
jeffmaury: the contextdir is relative to this file

rag-langchain/ai-studio.yaml
```diff
@@ -0,0 +1,22 @@
...
+      model-service: true
+      backend: 
+        - llama
+      arch:
+        - arm64
+        - amd64
+    - name: chromadb-server
+      image: docker.io/chromadb/chroma
+    - name: rag-inference-app
+      contextdir: rag-langchain
```
jeffmaury: Idem

jeffmaury (changes requested):

rag-langchain/ai-studio.yaml
MichaelClifford: Thanks @jeffmaury Let me know if the new change looks right.

rag-langchain/ai-studio.yaml
```diff
@@ -0,0 +1,22 @@
...
+      contextdir: ../playground
+      containerfile: Containerfile
+      model-service: true
+      backend: 
+        - llama
+      arch:
+        - arm64
+        - amd64
+    - name: chromadb-server
+      image: docker.io/chromadb/chroma
```
jeffmaury: image is not implemented yet. Replace by a single line containerfile

jeffmaury (changes requested):

rag-langchain/ai-studio.yaml
```diff
@@ -0,0 +1,22 @@
...
+    - name: llamacpp-server
+      contextdir: ../playground
+      containerfile: Containerfile
+      model-service: true
+      backend: 
+        - llama
+      arch:
+        - arm64
+        - amd64
+    - name: chromadb-server
```
lstocchi, line 15: We need a way to differentiate it from the sample app

lstocchi: Then idk how fast it is to start but if it's slow and the sample app fails to connect we need a way to listen to its port so when it's ready we can restart the sample app container. I need to test it

rag-langchain/rag_app.py
```diff
@@ -0,0 +1,89 @@
...
+from chromadb.config import Settings
+import chromadb.utils.embedding_functions as embedding_functions
+
+
+import uuid
+import os
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-d", "--docs", default="data/fake_meeting.txt")
```
lstocchi, line 20: Do we want to make it customizable?

rag-langchain/rag_app.py
```diff
@@ -0,0 +1,89 @@
...
+parser.add_argument("-c", "--chunk_size", default=150)
+parser.add_argument("-e", "--embedding_model", default="BAAI/bge-base-en-v1.5")
```
lstocchi, line 22: Do we want to make it customizable?

lstocchi (commented):
I'm ok with using the playground. My only concern is that we also need to have a playground-cuda if we're going to update the other samples

rag-langchain/ai-studio.yaml
lstocchi: We talked with @jeffmaury earlier. Now we find the port to listen to by using the containerfile (we look at the EXPOSE). If we don't have it as we use an image, maybe it's better to specify the port inside the ai-studio.yaml. if you decompile the image you could see multiple ports (also some belonging to the parent image)

rag-langchain/rag_app.py
MichaelClifford: Can you clarify? What would we want to customize here?

rag-langchain/rag_app.py
MichaelClifford: Can you clarify? What would we want to customize here?

rag-langchain/ai-studio.yaml
MichaelClifford: Isen't it differentiated by the `model-service=True` parameter?

sallyom: I agree - any information you need to build the application will be best if included in the ai.yaml - so you only have to depend on that single file

rag-langchain/rag_app.py
sallyom: it will be nice to make that a volume mounted/env var 
 value - so can use same image with arbitrary data

rag-langchain/ai-studio.yaml
sallyom: `FROM docker.io/chromadb/chroma`?

rag-langchain/rag_app.py
sallyom: replace with something like below so we can run podman with -e DOCS_PATH=wherever

```
default_docs = os.environ.get('DOCS_PATH', 'data/fake_meeting.txt')
parser.add_argument("-d", "--docs", default=default_docs)
```

rag-langchain/ai-studio.yaml
MichaelClifford: 👍 done

rag-langchain/ai-studio.yaml
MichaelClifford: wouldn't there be instances where you'd want the user to define the ports to use just before runtime?

rag-langchain/rag_app.py
MichaelClifford: That is basically how it works already. You can do `podman run ragapp -d <file-path>`. 
Is there a reason to use env vars over over command line parameters?  

Will this make exposing these options in the UI easier?

rag-langchain/ai-studio.yaml
lstocchi: so atm what we do is, if `model_service=true` we mount the volume/set he model_path else we set the model_endpoint env.
When running the chromadb service the model_endpoint is not really necessary, so it would be nice to know if the item we are handling is the sample app or the vector db.

Regarding the port, by reading the readme i see that we should run `podman run -it -p 8000:8000 chroma` but where do i take the port value i have to open if we do not add it in the ai-studio?

rag-langchain/rag_app.py
lstocchi: Bc we create a pod, add the containers (model_services, sample app, vectordb) in it and start the pod (then we restart the sample app again if it fails), so not sure if it's feasable to pass args this way. We can push a fix later btw.

rag-langchain/ai-studio.yaml
MichaelClifford: should we add a flag `vectordb=true`?

rag-langchain/rag_app.py
MichaelClifford: I'm happy to change it if it would be better, but its unclear to me why having them in podman run command as env variables instead of command line arguments would make a difference. The envs would still need to be passed to the command on starting the pod a second time.

MichaelClifford: o nvm, you just mean passing them this way at all into the pod might not work, is that right?

lstocchi: We pass the envs at creation so when we start, we just execute the start cmd

lstocchi: yes, not sure if with the pod works

lstocchi: but i'll make a test and let you know. Not urgent. it was just a thought

MichaelClifford: ok, I will change it.

rag-langchain/ai-studio.yaml
jeffmaury: we should keep in mind to be as much generic as we can. I don't see why we need to differentiate client app and vector db. If we need to pass the port to another container then we should do like what is done in Kubernetes ie have env var named CONTAINER_NAME_PORT

lstocchi: @jeffmaury i don't get your point. We need to know if we are working with a sample app or a vectorDB. When dealing with the sample app container we need to open its port into the pod so that it can be reached, nothing for the vectorDB. For the sample app we also need to fill the env variable so that it can connect to the model service, nothing for the vectorDB apparently. So there is a difference when adding a sample app or a vectorDB to the pod.

rag-langchain/rag_app.py
lstocchi: resolving as i think we can add cmds without using additional env

sallyom (approved):
lgtm

PR #32, opened 2024-01-30
add quadlet.image for chatbot

sallyom:
Adds an `.image` quadlet file to ensure application (chatbot) images are pulled on boot with bootc image
Example here: https://github.com/redhat-et/bootc-examples/blob/main/bootc-build/Containerfile


sallyom:
@MichaelClifford ptal, thank you!

PR #33, opened 2024-02-02
add chatbot langchain recipe

MichaelClifford:
This is a new chatbot recipe that should eventually replace the current `chatbot/` recipe. 

This demo utilizes Streamlit for the UI and Langchain to interface more generically with a model service. It has currently been tested agains the "playground" image model service.


sallyom (approved):
ship it! LGTM

PR #34, opened 2024-02-02
update UI to streamlit

MichaelClifford:
Replaced the command line interface with a streamlit app.


MichaelClifford:
It looks like this. 
Screenshot 2024-02-02 at 2 56 46 PM

I think I'm going to switch all the UI's to use this tool instead of gradio moving forward.

sallyom (approved):
LGTM

PR #36, opened 2024-02-03
add chatbot-langchain quadlet files

sallyom:
/cc @MichaelClifford 
Added quadlet files for the chatbot-langchain example


MichaelClifford (approved):
LGTM 🚀

PR #37, opened 2024-02-06
add document selection

MichaelClifford:
Adds a document selection dropdown sidebar to the RAG app. 

Selecting a document will repopulate the vector database and allow the user to chat with the model using the selected document as additional context.  

Screenshot 2024-02-06 at 11 00 20 AM

### DIFF
--- a/rag-langchain/rag_app.py
+++ b/rag-langchain/rag_app.py
@@ -18,7 +18,6 @@ import os
 import argparse
 
 parser = argparse.ArgumentParser()
-parser.add_argument("-d", "--docs", default="data/fake_meeting.txt")
 parser.add_argument("-c", "--chunk_size", default=150)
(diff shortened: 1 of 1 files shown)


sallyom (approved):
LGTM!!

rag-langchain/rag_app.py
```diff
@@ -27,20 +26,35 @@
...
 args = parser.parse_args()
 
+def clear_vdb():
+    global client
+    client.delete_collection(args.name)
+    print("clearing DB")
+
+def get_files():
+    file_list = os.listdir("data/")
+    file_list = [f for f in file_list if f.endswith(".txt")]
```
sallyom, line 36: if possible can we show any file in the data dir?

sallyom (approved):
As a follow-up, a check for a text file will be added rather than checking for the `.txt` in the filename.

PR #38, opened 2024-02-07
Code generation app demo 

suppathak:
- Added an app.py script to create a Streamlit web application for code-generation.

### DIFF
--- a/code-generation/codegen-app.py
+++ b/code-generation/codegen-app.py
@@ -24,7 +24,7 @@ llm = ChatOpenAI(base_url=model_service,
 # Define the Langchain chain
 prompt = ChatPromptTemplate.from_template("""You are an helpful code assistant that can help developer to code for a given {input}. 
                                           Generate the code block at first, and explain the code at the end.
-                                          If the {input} is not making sense, please ask more clarification.""")
+                                          If the {input} is not making sense, please ask for more clarification.""")
 chain = (
     {"input": RunnablePassthrough()}
     | prompt
--- a/code-generation/codegen-app.py
+++ b/code-generation/codegen-app.py
@@ -1,0 +1,51 @@
+from langchain_openai import ChatOpenAI
+from langchain_core.prompts import ChatPromptTemplate
+from langchain_core.runnables import RunnablePassthrough
+
+import streamlit as st
+
+import argparse
+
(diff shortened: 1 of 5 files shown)


code-generation/codegen-app.py
```diff
@@ -0,0 +1,40 @@
...
+from langchain_core.prompts import ChatPromptTemplate
+from langchain_core.runnables import RunnablePassthrough
+
+import streamlit as st
+
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-m", "--model_url", default="http://0.0.0.0:8001/v1")
+args = parser.parse_args()
```
MichaelClifford: Let's us environment variables here instead. Something like:

```
import os
model_service = os.getenv("MODEL_SERVIC_ENDPOINT", "http://localhost:8001/v1")
```

and then use this variable on line 22.

code-generation/codegen-app.py
```diff
@@ -0,0 +1,40 @@
...
+    st.session_state["messages"] = [{"role": "assistant", 
+                                     "content": "How can I help you?"}]
+    
+for msg in st.session_state.messages:
+    st.chat_message(msg["role"]).write(msg["content"])
+
+llm = ChatOpenAI(base_url=args.model_url, 
+                 api_key="EMPTY",
+                 streaming=True)
+
```
MichaelClifford: Why not use the `StreamlitCallbackHandler()` here? This will add streaming output to your app.

code-generation/codegen-app.py
```diff
@@ -0,0 +1,40 @@
...
+prompt = ChatPromptTemplate.from_template("""You are helpful AI assistant to write a code for a given {input}""")
```
MichaelClifford: Do you think we can improve this prompt at all? Have you researched any prompt templates that might be a better fit for code gen?

suppathak: Thank you Michael for your feedbacks. I will address them accordingly.
Regarding prompt, I have included the basic prompt templates here. Will definitely experiment with couple of others and update the code.

MichaelClifford (approved):
LGTM

lstocchi (approved):
Tested and works fine 🥳 LGTM!

PR #39, opened 2024-02-07
adds check for text files to rag app

MichaelClifford:
Adds an additional step prior to populating the list of available files and checks if they are text files. This approach tries to open the file and confirm its contents. It replaces the previous method of just checking for ".txt" filename extensions.


sallyom (approved):
LGTM!

PR #40, opened 2024-02-09
add langchain summarizer application

MichaelClifford:
This is a new summarization recipe that should eventually replace the current `summarization/` recipe.

This demo utilizes Streamlit for the UI and Langchain to interface more generically with a model service. It has currently been tested against the "playground" image model service.

Screenshot 2024-02-09 at 4 38 45 PM


sallyom (approved):
LGTM

PR #45, opened 2024-02-14
Add multimodal support

MichaelClifford:
This PR adds a new Multimodal recipe for image analysis:

* Updated the `playground/run.sh` to support using multimodal models. Happy to break this into a separate "multimodal playground" if we think that would be better.  
* Added an `image_analysis/` dir with an example application where a user can upload an image and get a natural language response about its contents.


sallyom (approved):
This looks great - I will add the image build to the GH actions workflow as well as a quadlet for this in a follow-up!

lstocchi (commented):
@MichaelClifford so if i just select one model it won't work?

@jeffmaury maybe we should add the info that it works with multi models somewhere in ai-studio

MichaelClifford:
@lstocchi I'm not sure I understand your question. This adds multimodal support not multimodel support. This means it allows for models that can take multiple modalities (text and image) as input.

lstocchi:
> @lstocchi I'm not sure I understand your question. This adds multimodal support not multimodel support. This means it allows for models that can take multiple modalities (text and image) as input.

I asked bc in the readme it's written `The multimodal models require two gguf files. Please download the following two files to `models/`, so there could be model services that accepts multiple models. Or is it true for any model services? Just asking after a quick look, i didn't test it.

if i only select one model, what happens? Just one  modality is enabled? Or it failed as 2 models are mandatory?

MichaelClifford:
It will work like it did before if no second gguf file is used. It just skips it if there is no second gguf.
`--clip_model_path ${CLIP_MODEL_PATH:=None} \`

jeffmaury (changes requested):
Don't run in AI Studio. The playground does not start

image

Maybe Windows related

MichaelClifford:
yes, I think the line breaks are probably causing this issue. I can move the whole thing to a single line.

MichaelClifford:
@jeffmaury, moved everything to a single line. 
I think this should work on Linux and Windows

image_understanding/image_understanding.py
```diff
@@ -0,0 +1,38 @@
+from openai import OpenAI
+import streamlit as st
+import base64
+import os
+
+model_service = os.getenv("MODEL_SERVICE_ENDPOINT",
+                          default="http://localhost:8001/v1")
```
leseb, line 7: Why fallback on localhost? Assuming everything runs inside a container this `localhost` will never be reachable, right? How about failing if `MODEL_SERVICE_ENDPOINT` is empty?

image_understanding/builds/Containerfile
```diff
@@ -0,0 +1,8 @@
+FROM registry.access.redhat.com/ubi9/python-39:latest
+WORKDIR /image_understanding
+COPY builds/requirements.txt .
+RUN pip install --upgrade pip
```
leseb:
```suggestion
RUN pip install --upgrade pip && \
  pip install --no-cache-dir --upgrade -r requirements.txt
```

leseb: You also don't need `/image_understanding/requirements.txt` since the WORKDIR is already set to `image_understanding`.

image_understanding/image_understanding.py
MichaelClifford: We could. However, I find this is fairly helpful while developing, before we've containerized anything.

image_understanding/builds/Containerfile
MichaelClifford: I like it. Thanks :)

image_understanding/builds/Containerfile
```diff
@@ -0,0 +1,9 @@
+FROM registry.access.redhat.com/ubi9/python-39:latest
+WORKDIR /image_understanding
+COPY builds/requirements.txt .
+RUN pip install --upgrade pip && \
+  pip install --no-cache-dir --upgrade -r requirements.txt
+RUN pip install --no-cache-dir --upgrade -r /image_understanding/requirements.txt
```
leseb: This line needs to go away now :)

MichaelClifford: lol, thanks @leseb fixed!

sallyom:
@MichaelClifford ready to merge? LGTM

MichaelClifford:
@jeffmaury Can we get your sign off to merge this?

jeffmaury (approved):
LGTM

PR #47, opened 2024-02-16
fix: fix rag-langchain app

lstocchi:
I was working on adding the rag app to ai-studio and i noticed some issues.

### DIFF
--- a/rag-langchain/README.md
+++ b/rag-langchain/README.md
@@ -30,5 +30,5 @@ snapshot_download(repo_id="BAAI/bge-base-en-v1.5",
 Follow the instructions below to build you container image and run it locally. 
 
 * `podman build -t ragapp rag-langchain -f rag-langchain/builds/Containerfile`
-* `podman run --rm -it -p 8501:8501 -v Local/path/to/locallm/data:/rag/data:Z ragapp -- -H 10.88.0.1 -m http://10.88.0.1:8001/v1`
+* `podman run --rm -it -p 8501:8501 -v Local/path/to/locallm/models/:/rag/models:Z -v Local/path/to/locallm/data:/rag/data:Z ragapp -- -H 10.88.0.1 -e MODEL_SERVICE_ENDPOINT=http://10.88.0.1:8001/v1`
 
--- a/rag-langchain/README.md
+++ b/rag-langchain/README.md
@@ -30,5 +30,5 @@ snapshot_download(repo_id="BAAI/bge-base-en-v1.5",
 Follow the instructions below to build you container image and run it locally. 
 
 * `podman build -t ragapp rag-langchain -f rag-langchain/builds/Containerfile`
-* `podman run --rm -it -p 8501:8501 -v Local/path/to/locallm/models/:/rag/models:Z -v Local/path/to/locallm/data:/rag/data:Z ragapp -- -H 10.88.0.1 -e MODEL_SERVICE_ENDPOINT=http://10.88.0.1:8001/v1`
+* `podman run --rm -it -p 8501:8501 -v Local/path/to/locallm/models/:/rag/models:Z -v Local/path/to/locallm/data:/rag/data:Z -e MODEL_SERVICE_ENDPOINT=http://10.88.0.1:8001/v1 ragapp -- -H 10.88.0.1 `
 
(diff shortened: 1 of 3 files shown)


rag-langchain/README.md
```diff
@@ -30,5 +30,5 @@ snapshot_download(repo_id="BAAI/bge-base-en-v1.5",
 Follow the instructions below to build you container image and run it locally. 
 
 * `podman build -t ragapp rag-langchain -f rag-langchain/builds/Containerfile`
-* `podman run --rm -it -p 8501:8501 -v Local/path/to/locallm/models/:/rag/models:Z -v Local/path/to/locallm/data:/rag/data:Z ragapp -- -H 10.88.0.1 -m http://10.88.0.1:8001/v1`
+* `podman run --rm -it -p 8501:8501 -v Local/path/to/locallm/data:/rag/data:Z ragapp -- -H 10.88.0.1 -m http://10.88.0.1:8001/v1`
```
MichaelClifford: without the `models/` volume mount you will download the embedding model at runtime. I think we want to avoid downloading models at runtime.

rag-langchain/rag_app.py
```diff
@@ -18,15 +18,22 @@
...
+                          "http://0.0.0.0:8001/v1")
+
 parser = argparse.ArgumentParser()
 parser.add_argument("-c", "--chunk_size", default=150)
 parser.add_argument("-e", "--embedding_model", default="BAAI/bge-base-en-v1.5")
 parser.add_argument("-H", "--vdb_host", default="0.0.0.0")
 parser.add_argument("-p", "--vdb_port", default="8000")
 parser.add_argument("-n", "--name", default="test_collection")
-parser.add_argument("-m", "--model_url", default="http://0.0.0.0:8001/v1")
+parser.add_argument("-m", "--model_url", default=model_service)
```
MichaelClifford: I think this line can actually just be removed. Its duplicate information as `model_service`

rag-langchain/README.md
lstocchi: I don't get the role of the model service then. Isn't this going to first look at the vector db and then call the model service which has the model mounted?

MichaelClifford: It is a separate model for creating the vector embeddings. RAG apps need two models: a language model, which is in the model service and an embedding model in the client to convert input text to vector for the vector db.

rag-langchain/rag_app.py
```diff
@@ -18,15 +18,22 @@
...
 args = parser.parse_args()
 
+client = HttpClient(host=args.vdb_host,
```
MichaelClifford: If we are going to move it this far up can we rename it to `vectorDB_client`?

lstocchi: Sure, i moved that there bc i was getting an error "client is not defined in line 32"

MichaelClifford: Yeah, I figured it must have been throwing an error. Strange I never noticed it. Thanks :)

rag-langchain/README.md
```diff
@@ -30,5 +30,5 @@ snapshot_download(repo_id="BAAI/bge-base-en-v1.5",
 Follow the instructions below to build you container image and run it locally. 
 
 * `podman build -t ragapp rag-langchain -f rag-langchain/builds/Containerfile`
-* `podman run --rm -it -p 8501:8501 -v Local/path/to/locallm/models/:/rag/models:Z -v Local/path/to/locallm/data:/rag/data:Z ragapp -- -H 10.88.0.1 -m http://10.88.0.1:8001/v1`
+* `podman run --rm -it -p 8501:8501 -v Local/path/to/locallm/models/:/rag/models:Z -v Local/path/to/locallm/data:/rag/data:Z ragapp -- -H 10.88.0.1 -e MODEL_SERVICE_ENDPOINT=http://10.88.0.1:8001/v1`
```
MichaelClifford:
```suggestion
* `podman run --rm -it -p 8501:8501 -v Local/path/to/locallm/models/:/rag/models:Z -v Local/path/to/locallm/data:/rag/data:Z -e MODEL_SERVICE_ENDPOINT=http://10.88.0.1:8001/v1 ragapp -- -H 10.88.0.1 `
```

the `-e` and environment variable need to be set before the image name.

lstocchi: nice catch. Thanks!

MichaelClifford (approved):
LGTM! 🚀 
Thanks for the fixes!

PR #51, opened 2024-02-19
feat: add ports to ai-studio

lstocchi:
This PR adds a new property `Ports` in the ai-studio.yaml files.
This includes the ports used by the every single element within an ai application (model_service, vectordb, ai app).

It will be used by ai-studio to know which port to use when opening an app in the browser or which ports to set in the pod (also if we split the model_services and the ai app in different pods in future).

This is needed bc we now retrieve the ports info from the images exposedPorts property but this can contain ports related to the base image which is something we do not need to handle. E.g. by using the  `registry.access.redhat.com/ubi9/python-39:latest` base image the port 8080 is also exposed.

### DIFF
--- a/chatbot-langchain/ai-studio.yaml
+++ b/chatbot-langchain/ai-studio.yaml
@@ -12,9 +12,13 @@ application:
       arch:
         - arm64
         - amd64
+      ports:
+        - 8001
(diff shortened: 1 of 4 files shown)


MichaelClifford (approved):
LGTM

PR #53, opened 2024-02-19
fix: add model service check to Summarizer

lstocchi:
Code taken  from https://github.com/redhat-et/locallm/pull/41 to update the summarizer.


MichaelClifford (approved):
LGTM! 🚀

PR #41, opened 2024-02-12
add model service check

MichaelClifford:
This PR adds a check to the application to make sure the model service is running before proceeding. Once this PR is approved, we can roll out this same approach on the rest of the AI applications.

PR #54, opened 2024-02-19
fix: add model service check to code-generation app

lstocchi:
code taken from #41 . This updates the code-generation app


MichaelClifford (approved):
LGTM

PR #55, opened 2024-02-19
Update image references from `quay.io/sallyom` to `quay.io/redhat-et` and update readme

sallyom:
This PR updates image references and documentation

/cc @MichaelClifford


MichaelClifford (approved):
LGTM

PR #57, opened 2024-02-21
Add model conversion tool

MichaelClifford:
This PR adds a new directory `convert_models/` to the repo. This new directory contains the artifacts needed to build an image that can be used to convert models from huggingface into `*.gguf` format needed for llamacpp and local usage. This PR also contains a simple streamlit app to provide a rough outline of how this tool could be used by or integrated into AI studio. 

The reason we need this PR is to provide the user with more options and control over which models they want to use, and removes the dependency on interim model providers. For example: currently, if we want to use Mistral we need to get a `*.gguf` version from a third party like https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF This has obvious issues for usage, licensing and security. This tool allows us to pull the original model directly from the model developers org and perform the model conversions ourselves.


MichaelClifford:
cc @slemeur @mairin

jeffmaury (approved):
LGTM

slemeur:
Very nice @MichaelClifford ! Thanks

sallyom (approved):
LGTM!

PR #58, opened 2024-02-22
Add whisper.cpp model

hemajv:
Adding the containerfile to build the whisper.cpp image

### DIFF
--- a/whisper-playground/Containerfile
+++ b/whisper-playground/Containerfile
@@ -1,0 +1,23 @@
+FROM registry.access.redhat.com/ubi9/ubi:latest as builder
+
+WORKDIR /app
+RUN dnf install -y git make gcc gcc-c++
+RUN mkdir whisper && cd whisper && git clone https://github.com/ggerganov/whisper.cpp.git . && \
+    git checkout tags/v1.5.4 && \
+    make && \
+    cp main /app/main && \
+    cp samples/jfk.wav /app/jfk.wav && \
+    cd ../ && rm -rf whisper
+
+FROM registry.access.redhat.com/ubi9-minimal:latest
+WORKDIR /app
+COPY --from=builder /app /app
+
+# https://github.com/wader/static-ffmpeg
+# https://hub.docker.com/r/mwader/static-ffmpeg/
+COPY --from=mwader/static-ffmpeg:6.1.1 /ffmpeg /bin/
+COPY --from=mwader/static-ffmpeg:6.1.1 /ffprobe /bin/
+
+COPY run.sh /app/
+ENV AUDIO_FILE=/app/jfk.wav
+ENTRYPOINT ["sh", "run.sh"]
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.gguf
+*.bin
 *_pycache_*
--- a/whisper-playground/README.md
+++ b/whisper-playground/README.md
@@ -1,0 +1,55 @@
+### Build Model Service
+
+From this directory,
+
+```bash
+podman build -t whisper:image .
+```
+
+### Download Model
+
+We need to download the model from HuggingFace. There are various Whisper models available which vary in size and can be found [here](https://huggingface.co/ggerganov/whisper.cpp). We will be using the `small` model which is about 466 MB.
+
+- **small**
+    - Download URL: [https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-small.bin](https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-small.bin)
+
+```bash
+cd ../models
+wget --no-config --quiet --show-progress -O ggml-small.bin <Download URL>
+cd ../
+```
+
+### Download audio files
+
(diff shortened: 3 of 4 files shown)


whisper-playground/Containerfile
```diff
@@ -0,0 +1,14 @@
+FROM quay.io/centos/centos:stream8
```
jeffmaury: Why not using UBI ?

whisper-playground/Containerfile
```diff
@@ -0,0 +1,14 @@
...
+WORKDIR /app
+RUN dnf install -y make gcc gcc-c++ git && \
+  rm -rf /var/cache/dnf/*
+RUN git clone https://github.com/hemajv/whisper.cpp.git . && \
+    make
+USER root
```
jeffmaury: Is that required ?

whisper-playground/Containerfile
```diff
@@ -0,0 +1,14 @@
+FROM quay.io/centos/centos:stream8
+WORKDIR /app
+RUN dnf install -y make gcc gcc-c++ git && \
+  rm -rf /var/cache/dnf/*
+RUN git clone https://github.com/hemajv/whisper.cpp.git . && \
```
jeffmaury: I think we should use builder images to lower to footprint in the final image

whisper-playground/Containerfile
```diff
@@ -0,0 +1,14 @@
...
+    make
+USER root
+RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm \
+        https://download1.rpmfusion.org/free/el/rpmfusion-free-release-8.noarch.rpm \ 
+        https://download1.rpmfusion.org/nonfree/el/rpmfusion-nonfree-release-8.noarch.rpm \
+        https://rpmfind.net/linux/centos/8-stream/PowerTools/aarch64/os/Packages/SDL2-2.0.10-2.el8.aarch64.rpm && \
```
jeffmaury: Seems arch related

whisper-playground/Containerfile
jeffmaury: Do we need a fork ?

whisper-playground/Containerfile
hemajv: Unfortunately, UBI did not support the `ffmpeg` package since it requires the SDL2 library as a dependency. Therefore had to use Centos instead. 

Related ffmpeg issue with UBI:
https://stackoverflow.com/questions/73597789/how-to-install-ffmpeg-on-ubi-docker-images

whisper-playground/Containerfile
hemajv: Yes, in order to install `epel` and `rpmfusion`, we need to be root

whisper-playground/Containerfile
hemajv: Yes, since I was running this on my Mac (arm64) these were required in order to install the `ffmpeg` package successfully

whisper-playground/Containerfile
hemajv: The reason I used my fork was so that it would be static and not have any issues if the upstream repo had minor changes that could break the image. We could build from the upstream repo as well, but need to monitor for any changes that occur.

whisper-playground/Containerfile
sallyom: we might be able to use a static ffmpeg binary? looking at that now - something like https://github.com/wader/static-ffmpeg

whisper-playground/Containerfile
sallyom: we can also use the 1.5.4 tag to pin the upstream repo

whisper-playground/Containerfile
hemajv: @sallyom yes I saw this too, can try and see if this works

whisper-playground/Containerfile
hemajv: @sallyom sure will do that 👍

sallyom: @hemajv ptal here: https://github.com/hemajv/locallm/pull/1 I think this works well, wdyt?

hemajv: @sallyom tysm, going to test this out and merge in a bit 👍

whisper-playground/Containerfile
hemajv: resolved this by installing the static ffmpeg binary instead

whisper-playground/Containerfile
hemajv: updated the Containerfile with Sally's suggested changes

hemajv:
@MichaelClifford @sallyom @jeffmaury updated the Containerfile to be ubi compatible by downloading the static ffmpeg binary, as suggested by Sally. Please do review the new changes.

sallyom (approved):
LGTM but now, I have a commit in here so @jeffmaury @MichaelClifford ptal & tag/merge if all is good!

jeffmaury (approved):
LGTM

MichaelClifford (approved):
LGTM
Thanks @hemajv ! 🚀

PR #59, opened 2024-02-26
chore: add version to ai-studio.yaml

lstocchi:
This PR adds a version tag to all ai-studio.yaml files.

From a desktop perspective, we could use the version to handle different ai-studio schemas in future release in case of some breaking changes.


MichaelClifford (approved):
LGTM

PR #72, opened 2024-03-04
Add option for config file in playground

MichaelClifford:
This PR updates the playground's `run.sh` to allow for the use of config files. Which, in tern allows for the server to load and unload multiple models without restarting the service.  

This feature was requested by a couple members of the AI-studio team. Please consider this PR as a way to test and evaluate if we want to move forward with this approach to loading models in the playground.

### DIFF
--- a/playground/README.md
+++ b/playground/README.md
@@ -68,5 +68,5 @@ podman run --rm -it -d \
         -p 8001:8001 \
         -v Local/path/to/locallm/models:/locallm/models:ro,Z \
         -e CONFIG_PATH=models/<config-filename> \
-        playground:image`
+        playground:image
 ```
--- a/playground/README.md
+++ b/playground/README.md
@@ -23,7 +23,9 @@ cd ../
 
 ### Deploy Model Service
 
-Deploy the LLM server and volume mount the model of choice.
+#### Single Model Service:
+
+Deploy the LLM server and volume mount the model of choice using the `MODEL_PATH` environment variable.
 
 ```bash
 podman run --rm -it -d \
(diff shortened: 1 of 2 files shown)


MichaelClifford:
cc @axel7083

playground/README.md
```diff
@@ -34,3 +36,37 @@ podman run --rm -it -d \
...
+}
+```
+
+Now run the container with the specified config file. 
+```bash
+podman run --rm -it -d \
+        -p 8001:8001 \
+        -v Local/path/to/locallm/models:/locallm/models:ro,Z \
+        -e CONFIG_PATH=models/<config-filename> \
+        playground:image`
```
leseb:
```suggestion
        playground:image
```

playground/run.sh
```diff
@@ -1,2 +1,14 @@
 #!/bin/bash
-python -m llama_cpp.server --model ${MODEL_PATH} --host ${HOST:=0.0.0.0} --port ${PORT:=8001} --n_gpu_layers 0
+
+if [ ${CONFIG_PATH} ] || [[ ${MODEL_PATH} && ${CONFIG_PATH} ]]; then
```
leseb, line 2: What do you mean by `[[ ${MODEL_PATH} && ${CONFIG_PATH} ]]`, if both are set you want to use `CONFIG_PATH`? IMO this case should be an error because the user has no clear idea what the behavior is going to be if both are given. How about the following instead?

```bash
if [[ -n "${CONFIG_PATH}" ]] && [[ -n "${MODEL_PATH}" ]]; then
  echo "Error: both CONFIG_PATH and MODEL_PATH cannot be set, choose only one"
  exit 1
fi

if [[ -n "${CONFIG_PATH}" ]]; then
    python -m llama_cpp.server --config_file "${CONFIG_PATH}"
elif [[ -n "${MODEL_PATH}" ]]; then
    python -m llama_cpp.server --model "${MODEL_PATH}" --host "${HOST:=0.0.0.0}" --port "${PORT:=8001}" --n_gpu_layers 0
else
    echo "Error: Either CONFIG_PATH or MODEL_PATH must be set."
    exit 1
fi
```

MichaelClifford: The current behavior of `llama_cpp.server` is that it will "work" if both parameters are provided, but it will default to the config if both were given. So my goal was to just pass the existing behavior through. 

Also, this is the main script for running the playground on AI studio, so I wanted to be as permissive as possible to not break anything existing . 

That said, I think your suggestion is probably more user friendly. Happy to implement if others think it will work fine for AI studio.

jeffmaury (commented):
Still having an issue with LF conversions. Can you rebase ?

MichaelClifford:
@jeffmaury rebased

MichaelClifford:
@jeffmaury @axel7083

Rebased and fixed merge conflicts. Please take a look.

jeffmaury (approved):
LGTM

PR #75, opened 2024-03-06
Add experimental llamacpp/vulkan server

slp:
Refactor playground to support multiple model servers, and implement a vulkan variant.

### DIFF
--- a/model_servers/README.md
+++ b/model_servers/README.md
@@ -1,0 +1,37 @@
+### Build Model Service
+
+From this directory,
+
+```bash
+cd llamacpp
+podman build -f base/Containerfile -t playground:image .
+```
+
+### Download Model
+
+At the time of this writing, 2 models are known to work with this service
+
+- **Llama2-7b**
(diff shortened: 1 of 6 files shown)


MichaelClifford:
Thanks for the PR @slp. I plan on depreciating this recipe soon in place of an updated one that relies on a shared model server. 

Could you add this Containerfile to a new directory instead `locallm/model_servers/llamacpp/vulkan/` ?

slp:
@MichaelClifford I've updated the PR to refactor `playground` as `model_servers`. PTAL.

MichaelClifford:
Could you please move these files to a new directory `model_servers/` and not update the `playground/` directory right now. [AI Studio](https://github.com/projectatomic/ai-studio) relies on the playground and I'd would prefer to no make too many changes there just yet.

MichaelClifford:
It might look slightly duplicative at the moment, but as we add additional model servers that may or may not be used by the playground, I'd like to have a clear place for them in the repo.

slp:
Sure, no problem, I'll update the PR.

slp:
Done. Please let me know if it needs more adjustments.

PR #76, opened 2024-03-06
Remove old recipes

MichaelClifford:
This PR removes the older versions of the recipes `chatbot`, `summarizer` and `rag`  and replaces them with updated versions that we're previously named `<RECIPE>-langchain`.

The reason that this PR is needed is that the earlier versions of recipes followed a pattern we no longer wish to support and leaving them in the repo was causing some confusion among users.


sallyom (changes requested):
I'd like to update the [.github/workflows/build-images.yaml](https://github.com/redhat-et/locallm/blob/main/.github/workflows/build-images.yaml) before merging this, to ensure all image builds are still good.

sallyom:
@MichaelClifford would it be possible to re-work this, so that there is 
* 1 PR that simply `rm -rf chatbot summarizer rag` 
* 1 PR to rename everything
This will be cleaner, also, I don't mind if we keep them names as/is - I like the descriptive names that include 'langchain'

MichaelClifford:
@sallyom Lol, I considered that at first but then decided to push it all as one big PR. 

Sure, let me re-do this one to just delete the older ones. Then I'll follow up witha re-name pr.

sallyom (approved):
LGTM!

PR #80, opened 2024-03-12
Add whisper streamlit app

hemajv:
Including streamlit front end application for interacting with the Whisper model


MichaelClifford (approved):
LGTM

PR #96, opened 2024-03-24
playground tests

lmilbaum:


lmilbaum:
The following test throws an error which I'm still debugging:
```
================================================================================ ERRORS =================================================================================
____________________________________________________________ ERROR collecting playground/tests/test_alive.py ____________________________________________________________
/opt/app-root/lib64/python3.11/site-packages/pluggy/_hooks.py:501: in __call__
    return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
/opt/app-root/lib64/python3.11/site-packages/pluggy/_manager.py:119: in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
/opt/app-root/lib64/python3.11/site-packages/_pytest/python.py:276: in pytest_pycollect_makeitem
    return list(collector._genfunctions(name, obj))
/opt/app-root/lib64/python3.11/site-packages/_pytest/python.py:489: in _genfunctions
    self.ihook.pytest_generate_tests.call_extra(methods, dict(metafunc=metafunc))
/opt/app-root/lib64/python3.11/site-packages/pluggy/_hooks.py:562: in call_extra
    return self._hookexec(self.name, hookimpls, kwargs, firstresult)
/opt/app-root/lib64/python3.11/site-packages/pluggy/_manager.py:119: in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
playground/tests/conftest.py:7: in pytest_generate_tests
    pytest_container.auto_container_parametrize(metafunc)
/opt/app-root/lib64/python3.11/site-packages/pytest_container/helpers.py:36: in auto_container_parametrize
    raise ValueError(
E   ValueError: The test function test_alive is using the auto_container fixture but the parent module is not setting the 'CONTAINER_IMAGES' variable
======================================================================== short test summary info ========================================================================
ERROR playground/tests/test_alive.py - ValueError: The test function test_alive is using the auto_container fixture but the parent module is not setting the 'CONTAINER_IMAGES' variable
```

playground/tests/test_alive.py
```diff
@@ -0,0 +1,5 @@
+import pytest_container
+
+
+def test_alive(auto_container: pytest_container.container.ContainerData, host):
```
dcermak: You need to define which container images this test acts upon by setting the file global variable `CONTAINER_IMAGES`. Setting it in `conftest.py` is not sufficient.

lmilbaum: Thank you for your feedback. Should I also duplicate the image definition?

dcermak: No, just import `TW` from `conftest.py`.

playground/tests/conftest.py
```diff
@@ -0,0 +1,6 @@
+import pytest_container
+
+TW = pytest_container.Container(url="containers-storage:localhost/playground",forwarded_ports=[pytest_container.PortForwarding(container_port=8001)])
```
dcermak: Why are you relying on a locally built container and not just adding the containrefile here? Assuming that you are using `.devcontainer/Contairefile`, you could achieve this as follows :crossed_fingers: :
```suggestion
TW = pytest_container.DerivedContainer(
    base=" registry.access.redhat.com/ubi9/python-311:1-52",
    containerfile="""
USER root

COPY playground/tests/requirements.txt .

RUN pip3 install -r requirements.txt && \
    dnf install -y podman buildah && \
    dnf clean all
""",
    forwarded_ports=[pytest_container.PortForwarding(container_port=8001)]
)
```

lmilbaum: This is the right Containerfile:- https://github.com/redhat-et/locallm/blob/main/playground/Containerfile
Duplicating the Containerfile code in two source files is not something I'd like to do.

dcermak: Sure, but now your tests have an additional (undocumented) prerequisite. Also, nothing prevents you from reading the file in your python code.

lmilbaum: Additional undocumented prerequisite?

dcermak: Your tests will fail unless the user ran `buildah bud -f Containerfile -t playground .` in the root directory beforehand. If you give pytest_container the containerfile, then it'll build it yourself and the user needn't make this additional step themselves

lmilbaum: I would prefer the build is taken care by github actions job before pytest is executed

lmilbaum: @dcermak The desired workflow (not working yet) https://github.com/redhat-et/locallm/actions/runs/8423835317/job/23066387994?pr=96

playground/tests/conftest.py
```diff
@@ -0,0 +1,10 @@
+import pytest_container
+import os
+
+TW = pytest_container.Container(
+    url=f"containers-storage:{os.environ['REGISTRY']}/{os.environ['IMAGE_NAME']}",
```
dcermak: The `containers-storage:` prefix implies that the image is already present in the local image store and will **not** be pulled from the registry. So unless you want to manually pull it from the registry, I would suggest to drop that prefix:

```suggestion
    url=f"{os.environ['REGISTRY']}/{os.environ['IMAGE_NAME']}",
```

playground/tests/conftest.py
dcermak: Probably caused by this: https://github.com/redhat-et/locallm/pull/96#discussion_r1538031367

playground/tests/conftest.py
lmilbaum: The image should be present in the local image store. I can't figure out why it is not there.

dcermak: Why should it be there? You build it with docker and then run the tests with podman. podman will not look into docker's store when the `containers-storage:` prefix is used (that's for consistency with upstream behavior).

Why do you want to use the locally build image at all cost? Imho it only complicates matters for you.

lmilbaum: @dcermak I'm starting to understand what you are talking about. :-) 
The idea is to build and test the image in CI before it is pushed to a registry.  AFAIK, it is impossible to push the image to ghcr.io from a forked repository PR. That is why I'm trying to pick it up locally. Does it make sense?

lmilbaum: Switched the build step to buildah and it looks better now.  Your insight helped me figure this out. Thank you.

dcermak: Happy to help, thanks for using pytest_container!

playground/Makefile
```diff
@@ -0,0 +1,3 @@
+.PHONY: build
+build:
+	buildah bud -f Containerfile --no-cache -t playground .
```
sallyom: @lmilbaum is `make build` called somewhere? I don't see it

.github/workflows/playground.yaml
```diff
@@ -0,0 +1,50 @@
...
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository_owner	}}/playground
```
sallyom: @MichaelClifford keep this model-server used in AI Lab playground as `playground`? and for other uses we'll name the model_server/llamacpp/base/Containerfile image `model-server-llamacpp` or something similar, sg?

sallyom (commented):
looking good - and once we merge this let's add workflows for _every image_!

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "ai_recepies",
```
sallyom: recipes

playground/Makefile
lmilbaum: It is not. I used it temporarily. Will remove it.

.devcontainer/Containerfile
```diff
@@ -0,0 +1,9 @@
+FROM registry.access.redhat.com/ubi9/python-311:1-52
+
+USER root
+
+COPY playground/tests/requirements.txt .
+
+RUN pip3 install -r requirements.txt && \
+    dnf install -y podman buildah && \
```
dcermak: podman and buildah don't work without additional modifications in a container. You have to switch to a different storage driver as the overlay driver is not available inside containers

playground/tests/test_alive.py
```diff
@@ -0,0 +1,11 @@
...
+from .conftest import TW
+
+CONTAINER_IMAGES = [TW]
+
+
+# def test_etc_os_release_present(auto_container: pytest_container.container.ContainerData):
+#     assert auto_container.connection.file("/etc/os-release1").exists
+
+def test_alive(auto_container: pytest_container.container.ContainerData, host):
+    res = host.run_expect([0],f"curl localhost:8001",).stdout.strip()
```
dcermak: I guess this fails because the container's entrypoint/cmd didn't launch yet. You don't see this in manual runs, because you're generally not as fast as the machine, which starts your container and immediately `curl`s the port. My recommendation is to use `HEALTHCHECK` to ensure that the application inside the container is up and running.

lmilbaum: Ahh... that might be a problem. Unfortunately, HEALTHCHECK is not an OCI compatible instruction.  I'll look for another option.

.devcontainer/Containerfile
lmilbaum: Thanks for your input. I'll have to consult with folks who have more knowledge and experience with it.

lmilbaum: It is probably not the problem. I've invoked the container and the curl command mainauly and it worked as expected.

lmilbaum: Found the problem. Unrelated to the plugin. Thanks.

playground/Makefile
```diff
@@ -0,0 +1,19 @@
+.PHONY: build
+build:
+	podman build -f playground/Containerfile -t ghcr.io/redhat-et/playground --format docker playground
+
+models/llama-2-7b-chat.Q5_K_S.gguf:
+	wget -P models https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGUF/resolve/main/llama-2-7b-chat.Q5_K_S.gguf -o /dev/null
+	# curl -s -S -L -f https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGUF/resolve/main/llama-2-7b-chat.Q5_K_S.gguf -z $@ -o $@.tmp && mv -f $@.tmp $@ 2>/dev/null || rm -f $@.tmp $@
```
cgwalters: I see this is implementing the "transactional write" pattern, does `wget` actually do that by default?  It doesn't seem so.  Curious why wget over curl.

playground/tests/conftest.py
```diff
@@ -0,0 +1,32 @@
+import pytest_container
+import os
+
+
+TW = pytest_container.Container(
```
cgwalters: What is "TW" doing here?

playground/Makefile
```diff
@@ -0,0 +1,19 @@
...
+
+.PHONY: install
+install:
+	pip install -r playground/tests/requirements.txt
```
cgwalters: Hm, why do we need to reinstall here?  Isn't that done in the container image?  Or are there separate things?

playground/tests/conftest.py
```diff
@@ -0,0 +1,32 @@
+import pytest_container
```
cgwalters: We chatted about this, I don't have extensive experience with testing "bare" app containers outside of Kube/OCP but I did want to just link for reference e.g. https://github.com/osbuild/bootc-image-builder/tree/main/test which is also using pytest to test a container image, though not using this library.

cgwalters (commented):
I think this makes sense.  Overall comment is it could use a brief high level architecture description (possibly in the Makefile?) or a separate readme.md.

playground/Makefile
lmilbaum: The curl command is preferred but it is broken on GHA. I'll try to fix it.

playground/tests/conftest.py
lmilbaum: This is where pytest-container plugin runs the playground container.  TW is just a "name".

playground/Makefile
lmilbaum: Those are separate things. This is the list of requirements the GHA runners should have to run pytest.

playground/tests/conftest.py
lmilbaum: Thanks. I'll check it out.

lmilbaum: It seems that the container functions were implemented instead of using the plugin.   Personally, I prefer using a plugin.

rhatdan:
@MichaelClifford PTAL

playground/tests/conftest.py
dcermak: I believe this is a remnant from one of the examples from `pytest_container`. The image I used in that example was "openSUSE Tumbleweed" which I just abbreviated with `TW`. But in the end it's just a name, albeit I think that `TW` is not very suitable here.

playground/tests/test_alive.py
```diff
@@ -0,0 +1,13 @@
...
+import time
+
+CONTAINER_IMAGES = [TW]
+
+
+def test_etc_os_release_present(auto_container: pytest_container.container.ContainerData):
+    assert auto_container.connection.file("/etc/os-release").exists
+
+def test_alive(auto_container: pytest_container.container.ContainerData, host):
+    time.sleep(10)
```
dcermak: This will give you a pretty hefty delay on each test. You can try something similar to this: https://github.com/SUSE/BCI-tests/blob/aa25debbe104f5c856edd54cd1d653a75bb4b60e/tests/test_tomcat.py#L36

lmilbaum: Thanks. That is a good idea.

playground/tests/conftest.py
lmilbaum: Ah, now I understand what TW stands for. I was asking myself this question.
Replaced it with something which makes more sense to this use case.

playground/README.md
```diff
@@ -20,6 +24,10 @@ cd ../models
 wget <Download URL>
 cd ../
 ```
+or
+```bash
+make -f Containerfile download
```
MichaelClifford: This seems like it downloads the requirements, not the model.

playground/Makefile
```diff
@@ -0,0 +1,18 @@
+.PHONY: build
+build:
+	podman build -f playground/Containerfile -t ghcr.io/ai-lab-recipes/playground --format docker playground
+
+models/llama-2-7b-chat.Q5_K_S.gguf:
```
MichaelClifford:
```suggestion
.PHONEY: download
download:
```
does it make sense to make this "download" instead of the model name?

playground/Makefile
```diff
@@ -0,0 +1,18 @@
...
+	curl -s -S -L -f https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGUF/resolve/main/llama-2-7b-chat.Q5_K_S.gguf -z $@ -o $@.tmp && mv -f $@.tmp $@ 2>/dev/null || rm -f $@.tmp $@
+
+.PHONY: download
+download:
```
MichaelClifford:
```suggestion
.PHONY: install
install:
```
Does it make sense to rename this "install" and add "download" above for the model?

playground/Makefile
```diff
@@ -0,0 +1,18 @@
...
+	pip install -r playground/tests/requirements.txt
+
+.PHONY: run
+run: install models/llama-2-7b-chat.Q5_K_S.gguf
```
MichaelClifford:
```suggestion
run: install download
```

playground/Makefile
```diff
@@ -0,0 +1,18 @@
...
+	podman run -it -d -p 8001:8001 -v ./models:/locallm/models:ro,Z -e MODEL_PATH=models/llama-2-7b-chat.Q5_K_S.gguf -e HOST=0.0.0.0 -e PORT=8001 --net=host ghcr.io/redhat-et/playground
+
+.PHONY: test
+test: models/llama-2-7b-chat.Q5_K_S.gguf download
```
MichaelClifford:
```suggestion
test: download install
```

playground/Makefile
MichaelClifford: Is there a way to check if the model exists before downloading it? We don't want people re-downloading the model each time the use `make run`

playground/Makefile
```diff
@@ -0,0 +1,18 @@
+.PHONY: build
+build:
+	podman build -f playground/Containerfile -t ghcr.io/ai-lab-recipes/playground --format docker playground
+
+models/llama-2-7b-chat.Q5_K_S.gguf:
+	curl -s -S -L -f https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGUF/resolve/main/llama-2-7b-chat.Q5_K_S.gguf -z $@ -o $@.tmp && mv -f $@.tmp $@ 2>/dev/null || rm -f $@.tmp $@
```
MichaelClifford: Can we use this instead? https://huggingface.co/TheBloke/Mistral-7B-v0.1-GGUF/resolve/main/mistral-7b-v0.1.Q2_K.gguf 
Its smaller model (3gb) and has an apache license.

playground/README.md
lmilbaum: Good catch. Fixing.

playground/Makefile
lmilbaum: Reverted

playground/Makefile
lmilbaum: Yes. It is possible. I would prefer implementing it in a followup PR if that is not a critical change.

playground/Makefile
lmilbaum: That is doable. But, I would prefer implementing it on a followup PR.

sallyom (approved):
LGTM!! Thank you @lmilbaum this is really great and so much work to keep updating with the repo restructuring!!

PR #98, opened 2024-03-26
update python base images

sallyom:
version of cmake in python39 is 3.20, but llama-cpp-python requires >3.21 
updating the base python image everywhere to keep consistency 

/cc @MichaelClifford 
/cc @lmilbaum 
/cc @jeffmaury


audio-to-text/client/Containerfile
```diff
@@ -1,8 +1,8 @@
-FROM registry.access.redhat.com/ubi9/python-39:latest
+FROM registry.access.redhat.com/ubi9/python-311:latest
```
lmilbaum: Is it possible to pin the version? Using the latest tag might cause some issues

lmilbaum: Same comment is applied to everywhere the latest tag is being used. Renovate can be onboarded and used for bumping the versions.

MichaelClifford:
Would it make sense to pin the version of `llama_cpp_python` that worked earlier instead changing the base image? Seems like updating the base image might have wide reaching affect throughout the project. 

Did we confirm all of the model servers and AI apps work with python 3.11?

rhatdan:
Needs a rebase.

sallyom:
I'm going to pin both llamacpp-python _and_ the python image to match what Liora has in her tests, sg?

rhatdan:
Fine with me.

audio-to-text/client/Containerfile
sallyom: I updated to pin all python versions

lmilbaum (approved):
LGTM

PR #110, opened 2024-03-28
fix some podman build commands & update playground -> model_servers

sallyom:
Fallout from the move

@MichaelClifford ptal


rhatdan:
LGTM

PR #113, opened 2024-03-28
remove old /src files

MichaelClifford:
Deleted the `/src` along with `llamacpp_utils.py` thats no longer being used.


rhatdan:
LGTM
Test is broken but not related to this PR.

Issue #89, opened 2024-03-21
Restructuring Proposal


MichaelClifford:
I would like to change the structure of this repo a bit to make it more intuitive and easier to maintain as it continues to grow. Below please find a sketch of the proposed structure. 

The main changes would be:

* `playground/` would be moved into `model_servers/` and renamed `llamacpp_python/` 
* A new directory called `recipes/` with N subdirectories for each recipe category. Right now that is `natural_language_processing/`, `computer_vision/`, `audio/` and `multimodal/`. 
* All existing recipes will be moved to the correct subdirectory. 
* A new directory called `VectorDBs/`. With the intent to add any new component types that may arise in the future in the same way to the root directory. 
* Move `chromadb/` into `VectorDBs/`

Please let me know what you all think of the proposed changes and how they might affect any dependencies on this repo. 

```
locallm:
|-- assets
    |-- image.1.png
    |-- image2.png
    |-- ...
|-- data
    |-- file.txt
    |-- ...
|-- models
    |-- Model1Folder
    |-- Model2Folder
    |-- Convert_Models
    |-- ...
|-- model_servers
    |-- llamacpp_python
        |-- base
        |-- cuda
        |-- vulkan
        |-- ....
    |-- ollama
        |-- base
        |-- ....
    |-- caikit
        |-- base
        |-- ....
    |-- WhisperCPP
        |-- base
        |-- ....
    |-- ...
|-- recipes
    |-- Natural Language
        |-- Chat
        |-- Summary
        |-- RAG
        |-- Fine Tune
        |-- ...
    |-- Computer Vision
        |-- Object Detection
        |-- ...
    |-- Audio
        |-- Transcription
        |-- ...
    |-- Multimodal
        |-- Image Description
        |-- ...
|-- VectorDBs
    |-- ChromaDB
    |-- ...
```

sallyom:
I love this plan, I'm ok with it as soon as AI lab can sign off

jeffmaury:
What is the schedule ?

MichaelClifford:
@jeffmaury The sooner that better. When will it have the least impact on AI Studio? After 0.3.0?

jeffmaury:
This could be independant as the catalog is using commit SHAs

MichaelClifford:
@jeffmaury so we can start to make these changes today? Happy to wait until 0.3.0 to make things simple.

PR #115, opened 2024-03-28
Move `model_converter/`

MichaelClifford:
partially address #89

This PR moves `model_converter/` and associated code out of the root dir and into `models/`


rhatdan:
LGTM

PR #123, opened 2024-03-29
update mistral model to K_M and image updates to ai-lab

sallyom:
@rhatdan, I believe this is causing the error you're seeing `K_S` vs `K_M`


rhatdan:
LGTM

PR #124, opened 2024-03-30
Cleanup Makefile for chatbot

rhatdan:


recipes/natural_language_processing/chatbot/Makefile
```diff
@@ -1,19 +1,19 @@
...
+SERVERIMAGE ?= quay.io/ai-lab/llamacpp-python:latest
+SSHPUBKEY ?= $(shell cat ${HOME}/.ssh/id_rsa.pub;)
 .PHONY: build
 build:
 	podman build -f builds/Containerfile -t ghcr.io/ai-lab-recipes/${APP} .
 
 .PHONY: bootc
 bootc:
-	podman build --cap-add SYS_ADMIN --build-arg "SSHPUBKEY=$(SSHPUBKEY)" -f bootc/Containerfile -t ghcr.io/ai-lab-recipes/${APP}-bootc .
+	sudo podman build --cap-add SYS_ADMIN --build-arg "SSHPUBKEY=$(SSHPUBKEY)" -f bootc/Containerfile -t ghcr.io/ai-lab-recipes/${APP}-bootc .
```
cgwalters: It's tricky though because then this `Makefile` won't do the right thing on MacOS (and I think we want that right?)

So this goes back to using podman-machine even on Linux right?

sallyom: I think this will work on MacOS - If i'm running rootful podman machine I can run `podman` or `sudo podman` they both work.

sallyom: however, on my Mac with podman v5 - I need to add `--network=host` in order to build this.

sallyom: and also, on Mac, I cannot run nested podman cmds during podman build

rhatdan: I removed the sudo.

rhatdan: What it really means is requiring users to use rootful containers or figuring out why this is not working on rootless  mode.

PR #125, opened 2024-03-30
Add quadlet support for rag

rhatdan:


recipes/natural_language_processing/rag/bootc/Containerfile
```diff
@@ -0,0 +1,40 @@
...
+# public key into the image, allowing root access via ssh.
+
+FROM quay.io/centos-bootc/centos-bootc:stream9
+ARG SSHPUBKEY
+
+RUN mkdir /usr/etc-system && \
+    echo 'AuthorizedKeysFile /usr/etc-system/%u.keys' >> /etc/ssh/sshd_config.d/30-auth-system.conf && \
+    echo ${SSHPUBKEY} > /usr/etc-system/root.keys && chmod 0600 /usr/etc-system/root.keys
+
+RUN dnf -y update && (dnf install -y vim || true) && dnf clean all
```
cgwalters: This is probably more like `rpm -q vim || dnf install -y vim` right?

recipes/natural_language_processing/rag/bootc/Containerfile
```diff
@@ -0,0 +1,40 @@
...
+#
+# Use build command:
+# podman build --build-arg "sshpubkey=$(cat $HOME/.ssh/mykey.pub)" -t quay.io/exampleos/rag .
+# The --build-arg "SSHPUBKEY=$(cat ~/.ssh/mykey.pub)" option inserts your
+# public key into the image, allowing root access via ssh.
+
+FROM quay.io/centos-bootc/centos-bootc:stream9
+ARG SSHPUBKEY
+
+RUN mkdir /usr/etc-system && \
```
cgwalters, line 12: Yeah, I think we're going to want to do something like https://github.com/cgwalters/osbuild-cfg to avoid the copy-paste effect.  (not a blocker, just noting)

recipes/natural_language_processing/rag/bootc/README.md
```diff
@@ -0,0 +1,88 @@
...
+If not pre-pulling the workload images, you can cross build (ie, build from a Mac for an X86_64 system).
+To build the derived bootc image for x86_64 architecture, run the following:
+
+```bash
+cd recipes/natural_language_processing/rag
+
+# for CPU powered sample LLM application
+# to switch to aarch64 platform, pass --platform linux/arm64
+podman build --build-arg "sshpubkey=$(cat ~/.ssh/id_rsa.pub)" \
+           --cap-add SYS_ADMIN \
```
cgwalters: Let's add a comment that this is for the pre-pulling feature?

recipes/natural_language_processing/rag/Makefile
```diff
@@ -0,0 +1,39 @@
...
+	sudo podman build --cap-add SYS_ADMIN --build-arg "SSHPUBKEY=$(SSHPUBKEY)" -f bootc/Containerfile -t ghcr.io/ai-lab-recipes/${APP}-bootc .
+
+.PHONY: quadlet
+quadlet:
+# Modify quadlet files to match the server, model and app image
+	sed -e "s|SERVERIMAGE|${SERVERIMAGE}|" \
+	    -e "s|APPIMAGE|${APPIMAGE}|g" \
+	    -e "s|MODELIMAGE|${MODELIMAGE}|g" \
+	    quadlet/${APP}.image \
+	    > /tmp/${APP}.image
```
cgwalters: Hmm, what's up with the use of `/tmp` here?  Predictable filenames in `/tmp` are not good (https://cwe.mitre.org/data/definitions/377.html).

cgwalters (commented):
Generally looks sane to me!  Just some nits

recipes/natural_language_processing/rag/Makefile
rhatdan: Good point, I will clean this up.

recipes/natural_language_processing/rag/bootc/Containerfile
cgwalters: (Also a declarative package interface would handle this...another motivation for that)

rhatdan: I think we should remove this  no need for VIM in here.  Sally wanted it to make debugging easier, but I would just add VIM as an update to the image, if users want it.

sallyom: sgtm, remove it

recipes/natural_language_processing/rag/bootc/Containerfile
sallyom: i like it - can we just use this (everywhere?) ?

sallyom (changes requested):
for the RAG application, a vector database is added, so we'll need to add that into the pod def & quadlet & images, also.

rhatdan:
@sallyom Lets continue to iterate on this.

PR #126, opened 2024-03-31
Spin Model Server in chatbot recipe

lmilbaum:
1. As long as the repository source tree is dynamic, it would be preferable to invoke the functional tests on every change. 
2. Spin the model server container when testing chatbot app


rhatdan:
LGTM

Issue #67, opened 2024-02-28
add memory to rag and chat applications


MichaelClifford:
Both `chat_langchain` and `rag_langchain` do not have memory components. Meaning, that you cannot ask follow up questions during a chat session. 

We should add the a chat memory component that allows for asking follow up questions. It should also gracefully handle the case in which the chat history exceeds the models context window.

Gregory-Pereira:
Closing this issue as it was resolved in #127.

PR #127, opened 2024-03-31
Add memory to chatbot

MichaelClifford:
Partially addresses #67 

This updates `chatbot_ui.py` to include conversation memory to the chatbot, so that users can ask follow up questions.


rhatdan (approved):
LGTM

PR #132, opened 2024-04-01
updating readmes and makefiles for model_servers

Gregory-Pereira:
/cc @sallyom 
/cc @cooktheryan
/cc @MichaelClifford 

These changes have been tested on Darwin


Gregory-Pereira:
I noticed the code-generation llm was running extreemly slow, it took me about 5 minutes to get a response on  getting a python script with the openshift-client module to get all configmaps in ns `test` running on 128 gigs of RAM. I would consider that unsuable for your average person.

rhatdan:
@Gregory-Pereira This needs a rebase.

.gitignore
```diff
@@ -2,3 +2,6 @@
 *.bin
 *_pycache_*
 port_check.lock
+recipes/natural_language_processing/chatbot/chromedriver*
+models/.locks/
+models/models--BAAI--bge-base-en-v1.5/
```
MichaelClifford: I'm not sure if this is a sustainable approach since you'd have to add every model that gets put into your `models/` dir, right?. I would suggest adding the the entire directory to the .gitignore, but that will likely mean moving the `model_converter` out of that directory.

rhatdan: I agre, lets keep models separate from model_converter to separate directory.

rhatdan:
Needs a rebase.

.gitignore
Gregory-Pereira: Rebase has been done, what about this implementation: https://github.com/containers/ai-lab-recipes/pull/132/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R6-R9?
These lines will ignore everything in `models` but explicitly un-ignore `models/Containerfile`, `models/README.md`, and the `models/convert_models/` directory.

Gregory-Pereira:
I think tackling everything at once is not efficient. I will break this into 1 pr per componnt

Gregory-Pereira:
Didnt get to splitting this today, still have been slammed with productization work. I still plan to seperate these changes per component.

rhatdan:
LGTM

MichaelClifford:
@Gregory-Pereira should we close this PR in favor of the upcoming smaller ones?

Gregory-Pereira:
@MichaelClifford Yes, Ive just been held up by productization work. Happy to close this and il get to it whenever i can

PR #139, opened 2024-04-02
remove symlink to requirements.txt

sallyom:
removing the symlinks to requirements-test.txt and renaming to tests/requirements.txt

### DIFF
--- a/model_servers/llamacpp_python/tests/requirements.txt
+++ b/model_servers/llamacpp_python/tests/requirements.txt
@@ -1,0 +1,8 @@
+pip==24.0
+pytest-container==0.4.0
+pytest-selenium==4.1.0
+pytest-testinfra==10.1.0
+pytest==8.1.1
+requests==2.31.0
+selenium==4.19.0
+tenacity==8.2.3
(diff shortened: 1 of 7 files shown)


Gregory-Pereira (approved):
/lgtm

model_servers/llamacpp_python/tests/requirements.txt
```diff
@@ -0,0 +1,8 @@
+pip==24.0
```
rhatdan, line 1: Do we care about specific versions?  IE Will this become out of date over time?  Or do we plan on using renovate?

rhatdan:
LGTM

model_servers/llamacpp_python/tests/requirements.txt
Gregory-Pereira: I think its better to have all deps version locked, it will ensure that demos will be more robust. Ideally we also find a way to manage dependency updates like renovate or dependabot but for now I think this is fine

lmilbaum: renovate is planned to be onboarded

PR #142, opened 2024-04-02
move chatbot image builds

sallyom:
@lmilbaum for now, I've added a 2nd workflow that runs with cron schedule & trigger - this is so I can start updating the quay.io/ai-lab images as we go (so I don't have to build them manually). Eventually, the quay.io/ai-lab pushes will happen in the testing-framework workflow.

### DIFF
--- a/.github/workflows/chatbot.yaml
+++ b/.github/workflows/chatbot.yaml
@@ -32,12 +32,19 @@ jobs:
     steps:
       - uses: actions/checkout@v4.1.1
 
+      # required for multi-arch builds
+      - name: Install qemu dependency
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y qemu-user-static
+
       - name: Build Image
         id: build_image
         uses: redhat-actions/buildah-build@v2.13
         with:
           image: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: latest
+          platforms: linux/amd64, linux/arm64
           containerfiles: ./recipes/natural_language_processing/chatbot/builds/Containerfile
           context: recipes/natural_language_processing/chatbot
 
@@ -54,7 +61,7 @@ jobs:
         working-directory: ./recipes/natural_language_processing/chatbot
         run: make test
 
-      - name: Login to container registry
+      - name: Login to github container registry
         if: >
           (github.event_name == 'push' || github.event_name == 'schedule') &&
           (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v'))
--- a/.github/workflows/ailab-images.yaml
+++ b/.github/workflows/ailab-images.yaml
@@ -1,0 +1,55 @@
+name: Update quay.io/ai-lab images
+
+on:
+  schedule: # schedule the job to run at 12 AM daily
+   - cron: '0 0 * * *'
+  
+  workflow_dispatch:
+
+
+env:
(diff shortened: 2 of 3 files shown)


.github/workflows/chatbot.yaml
```diff
@@ -74,3 +85,48 @@ jobs:
           image: ${{ steps.build_image.outputs.image }}
           tags: ${{ steps.build_image.outputs.tags }}
           registry: ${{ env.REGISTRY }}
+
+      - name: Get changed chatbot files
```
lmilbaum: What if the workflow file changes? Or, could there be other files on top of the chatbot folder?

.github/workflows/chatbot.yaml
```diff
@@ -14,12 +14,16 @@ on:
     #   - ./recipes/natural_language_processing/chatbot/**
     #   - .github/workflows/chatbot.yaml
 
+permissions: write-all
```
lmilbaum: Is there a need for this to be set?

.github/workflows/chatbot.yaml
```diff
@@ -14,12 +14,16 @@ on:
...
+
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAME: ${{ github.repository_owner }}/chatbot
+  TEST_IMAGE_NAME: ${{ github.repository_owner }}/chatbot
+  QUAY: quay.io
```
lmilbaum: The images will be promoted to quay from the testing-framework workflow.

lmilbaum (changes requested):

.github/workflows/chatbot.yaml
sallyom: eventually, we'll gate running this with the `paths:` field - it does the same as this changed_files step, however, it will gate _all_ steps on this - with this, I wanted to only run the extra build & push on merge events. - I'll add to the changed files `.github/workflows/chatbot.yaml`

.github/workflows/chatbot.yaml
sallyom: let me remove it and see - I think I placed it there in the hopes of using cosign to sign the images based on a workload identity - we can add it back later if needed.

.github/workflows/chatbot.yaml
sallyom: oh ok! did you already add that? this is the only reason I'm adding anything here :) - let's see what's there, perhaps this PR can go away then!

sallyom: I'd like to start pushing the chatbot & llamacpp-python image in quay.io/ai-lab now with merges - since I'm currently manually building & pushing from local

sallyom: @lmilbaum I see what needs to happen here, I'll update this :)

.github/workflows/chatbot.yaml
sallyom: @lmilbaum I've moved the quay.io image build/pushes to the testing-framework workflow, ptal

.github/workflows/chatbot.yaml
sallyom: removed it

.github/workflows/chatbot.yaml
lmilbaum: How about adding a step to the testing-framework workflow to promote the images to quay.io?

lmilbaum (approved):

PR #143, opened 2024-04-02
Don't install .image quadlet in bootc Containerfiles

rhatdan:


cgwalters (approved):

recipes/natural_language_processing/chatbot/bootc/Containerfile
```diff
@@ -24,19 +24,32 @@ ARG APPIMAGE=quay.io/ai-lab/${RECIPE}:latest
...
+# Because images are prepulled, no need for .image quadlet
+# COPY quadlet/${RECIPE}.image /usr/share/containers/systemd
+# RUN sed -e "s|SERVERIMAGE|${SERVERIMAGE}|" \
+#    -e "s|APPIMAGE|${APPIMAGE}|g" \
+#    -e "s|MODELIMAGE|${MODELIMAGE}|g" \
+#    -i \
+#    /usr/share/containers/systemd/${RECIPE}.image
+
+# Setup /usr/lib/containers/storage as an additional store for images.
+# Remove once the base images have this set by default.
```
sallyom, line 45: the .image file ensures the images exist when restarting the service - if an image is manually deleted somehow then the service won't fail it will show as activating when pulling the image - is this enough of a reason to include it?  Also, it's important if you _don't_ pre-load the images.

recipes/natural_language_processing/chatbot/bootc/Containerfile
```diff
@@ -24,19 +24,32 @@ ARG APPIMAGE=quay.io/ai-lab/${RECIPE}:latest
...
 RUN sed -i -e '/additionalimage.*/a "/usr/lib/containers/storage",' \
         /etc/containers/storage.conf
 
+# Prepull the model, model_server and application to prepopulat the system.
```
sallyom: prepopulat  -> populate

recipes/natural_language_processing/chatbot/bootc/Containerfile
```diff
@@ -24,19 +24,32 @@ ARG APPIMAGE=quay.io/ai-lab/${RECIPE}:latest
...
 RUN podman pull --root /usr/lib/containers/storage ${SERVERIMAGE}
 RUN podman pull --root /usr/lib/containers/storage ${APPIMAGE}
 RUN podman pull --root /usr/lib/containers/storage ${MODELIMAGE}
+
+# Added for running as an OCI Container to prevent Overlay on Overlay issues.
```
cgwalters: Maybe let's link to https://github.com/CentOS/centos-bootc/issues/282 because there was some good discussion there about this; it's a complex topic.  Longer term as we go down the composefs path, maybe we can go to using cfs-fuse for the container-in-container case?

recipes/natural_language_processing/chatbot/Makefile
```diff
@@ -24,7 +24,7 @@ quadlet:
 	    -e "s|APPIMAGE|${APPIMAGE}|g" \
 	    -e "s|MODELIMAGE|${MODELIMAGE}|g" \
 	    quadlet/${APP}.yaml \
-	    > /tmp/${APP}.yaml
+	    > build/${APP}.yaml
```
sallyom, line 29: Maybe `mkdir -p .build` instead (and change L22,L27 to `.build/${APP}.*`) ? and then in .gitignore add this:
\*.build\*

recipes/natural_language_processing/chatbot/bootc/Containerfile
rhatdan: The problem is a image always attempts to pull, and no way to tell it to check if the image pre-exists.  Without the .image file the chatbot.yaml file should attempt to pull the image if it does not exists.

Since some of the images are protected behind a registry, we would need to populate a auth.json file, which would be a security risk or require users to setup the auth.json file some other way.  Bottom line, if  you prepull the images, then no need for a .image quadlet, as far as I can see.

recipes/natural_language_processing/chatbot/Makefile
rhatdan: .build would make it harder for users to find?  Why obscure this?

PR #146, opened 2024-04-02
Provisioning with user_data

lmilbaum:
provision.sh script is used to provision the instance that the terraform is spinning in the up step.  The script is consumed by the terraform configuration as the instance user_data.
the `aws ec2 wait` command is used for waiting until the provision.sh script is completed.
The destroy step which is invoked even if previous steps failed, is destroying the resources generated by the terraform configuration.


cgwalters (approved):
Looks sane to me

PR #147, opened 2024-04-03
add llamacpp-python & mistral to quay.io/ai-lab workflow

sallyom:
adds build & push for quay.io/ai-lab/llamacpp-python manifest for arm64 & amd64


sallyom:
this is only affecting image builds/pushes to quay.io/ai-lab - self-merging

PR #148, opened 2024-04-03
fix image name in ai-lab workflow

sallyom:
fix image references in ai-lab workflow


sallyom:
self-merging for quay.io/ai-lab build & push

PR #150, opened 2024-04-03
Separate model workflow

sallyom:
This
* enables triggering quay.io/ai-lab application & model-server build & push images separately from the quay.io/ai-lab model images
* removes the extra tag for the model image (doubles the storage) in favor of an image LABEL that describes the model file.


sallyom:
this will conclude the self-merges for quay.io/ai-lab build/push images :)

PR #156, opened 2024-04-03
update rag containerfile

sallyom:
updating rag Containerfile to remove the chromadb pieces that now run in a separate container.


rhatdan:
LGTM

PR #158, opened 2024-04-04
restore rag application Containerfile & other fixes 

sallyom:
* adds sqlite install back to rag containerfile
* adds data volume mount to rag pod definition

### DIFF
--- a/recipes/natural_language_processing/rag/builds/Containerfile
+++ b/recipes/natural_language_processing/rag/builds/Containerfile
@@ -1,4 +1,16 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-52
+### Update sqlite for chroma
+USER root
+RUN dnf remove sqlite3 -y
+RUN  wget https://www.sqlite.org/2023/sqlite-autoconf-3410200.tar.gz
+RUN tar -xvzf sqlite-autoconf-3410200.tar.gz
+WORKDIR sqlite-autoconf-3410200
+RUN ./configure
+RUN make
+RUN make install
+RUN mv /usr/local/bin/sqlite3 /usr/bin/sqlite3
+ENV LD_LIBRARY_PATH="/usr/local/lib"
+####
 WORKDIR /rag
 COPY builds/requirements.txt .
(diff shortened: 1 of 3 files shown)


recipes/natural_language_processing/rag/builds/Containerfile
```diff
@@ -1,4 +1,16 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-52
```
rhatdan, line 1: Why isn't this latest?

recipes/natural_language_processing/rag/builds/Containerfile
```diff
@@ -1,4 +1,16 @@
...
+### Update sqlite for chroma
+USER root
+RUN dnf remove sqlite3 -y
```
rhatdan, line 4: Is the sqlite package in RHEL not new enough?

 rpm -q sqlite
sqlite-3.34.1-7.el9_3.x86_64

recipes/natural_language_processing/rag/builds/Containerfile
sallyom: we've pinned to this everywhere - @lmilbaum are you ok w/ switching to latest everywhere?

recipes/natural_language_processing/rag/builds/Containerfile
sallyom: application errored without this added back, with this error: 
`Your system has an unsupported version of sqlite3. Chroma requires sqlite3 >= 3.35.0`
so i guess we just missed the cut there :)

recipes/natural_language_processing/rag/builds/Containerfile
rhatdan: My fear is that these do not age well.

recipes/natural_language_processing/rag/builds/Containerfile
rhatdan: Ok, hopefully 9.4 has an update.

MichaelClifford (approved):
lgtm

PR #160, opened 2024-04-04
Add bootc support for summarizer

rhatdan:


recipes/natural_language_processing/summarizer/Makefile
```diff
@@ -0,0 +1,50 @@
...
+	    > build/${APP}.image
+	sed -e "s|SERVERIMAGE|${SERVERIMAGE}|" \
+	    -e "s|APPIMAGE|${APPIMAGE}|g" \
+	    -e "s|MODELIMAGE|${MODELIMAGE}|g" \
+	    quadlet/${APP}.yaml \
+	    > build/${APP}.yaml
+	cp quadlet/${APP}.kube build/${APP}.kube
+
+.PHONY: install
+install:
```
MichaelClifford, line 33: I assume this is the same in the chatbot and rag, but do we want `make install` to do all this? Seems specific for the testing framework. 

Can we break it into "install" for the `requirements.txt` and something like "install-ci-test-reqs" for the rest? 

Not going to block on it here, but seems like we should update this throughout the repo.

MichaelClifford (approved):
LGTM

recipes/natural_language_processing/summarizer/Makefile
rhatdan: I don't know, I did not write that section.

PR #162, opened 2024-04-04
update rag bootc

sallyom:
fixing the bootc Containerfile for RAG
* adds a data directory on the host - also switch to using the `make quadlet` a dependent target for `make bootc` rather than adding the `sed` commands in the Containerfile
* updates the `build` to `bootc/build` so that the updated quadlet files can be copied in the bootc/Containerfile

TODO: document the `make` targets for each sample app

FOLLOWUP: move the `build` to `bootc/build` for `make quadlet` for other applications and update their bootc/Containerfiles


recipes/natural_language_processing/rag/Makefile
```diff
@@ -12,27 +12,28 @@ build:
 	podman build -f builds/Containerfile -t ${APPIMAGE} .
 
 .PHONY: bootc
-bootc:
-	podman build --cap-add SYS_ADMIN --build-arg "SSHPUBKEY=$(SSHPUBKEY)" -f bootc/Containerfile -t ${BOOTCIMAGE} .
+bootc: quadlet
+	sed -i '' -e "s|$(HOSTDATADIR)|/usr/share/rag/data|g" bootc/build/${APP}.yaml
+	podman build --cap-add SYS_ADMIN --build-arg "HOSTDATADIR=${HOSTDATADIR}" --build-arg "SSHPUBKEY=$(SSHPUBKEY)" -f bootc/Containerfile -t ${BOOTCIMAGE} .
```
rhatdan, line 17: I will merge this for now, but I believe that HOSTDATA should not be preloaded, so we might need to change what we are doing here.

recipes/natural_language_processing/rag/Makefile
```diff
@@ -12,27 +12,28 @@ build:
 	podman build -f builds/Containerfile -t ${APPIMAGE} .
 
 .PHONY: bootc
-bootc:
-	podman build --cap-add SYS_ADMIN --build-arg "SSHPUBKEY=$(SSHPUBKEY)" -f bootc/Containerfile -t ${BOOTCIMAGE} .
+bootc: quadlet
```
rhatdan, line 15: Nice change.

rhatdan:
LGTM

PR #168, opened 2024-04-05
model_service reorg to match docs

MichaelClifford:
Our docs are specific about how the `model_servers/` directories should be named. Updated them so that they all have consistent names and formats.


rhatdan:
LGTM

PR #169, opened 2024-04-05
update make bootc for chatbot, summarizer

sallyom:
@rhatdan updates `make bootc` to include `make quadlet` and moves generated quadlet files from ./build to ./bootc/build to be picked up by bootc/Containerfile

### DIFF
--- a/recipes/natural_language_processing/chatbot/Makefile
+++ b/recipes/natural_language_processing/chatbot/Makefile
@@ -10,7 +10,7 @@ build:
 	podman build -f builds/Containerfile -t ${APPIMAGE} .
 
 .PHONY: bootc
-bootc:
+bootc: quadlet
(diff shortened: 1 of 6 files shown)


recipes/natural_language_processing/chatbot/Makefile
```diff
@@ -21,13 +21,13 @@ quadlet:
 	    -e "s|APPIMAGE|${APPIMAGE}|g" \
 	    -e "s|MODELIMAGE|${MODELIMAGE}|g" \
 	    quadlet/${APP}.image \
-	    > build/${APP}.image
+	    > bootc/build/${APP}.image
```
rhatdan: Why not leave this at top level build directory and just change bootc/Containerfile to grab from build rather then quadlet?

I don't think make quadlet should be tied to bootc.

sallyom: updated

rhatdan:
LGTM
Merge once tests pass.

PR #171, opened 2024-04-05
initial docs for make commands

### LINKED ISSUE #164: Add documentation for `make` targets
We've added convenient `make` targets & we need to document them!

sallyom:
@rhatdan
@MichaelClifford 

These READMEs can be templated in the future, but for now I've duplicated information 

Resolves https://github.com/containers/ai-lab-recipes/issues/164


sallyom:
There are a few `podman build` and `podman run` commands that should also be replaced with their `make build/run` targets, I didn't test those though so didn't replace them here.

rhatdan:
LGTM

PR #172, opened 2024-04-05
WIP containerfile no contaienrs storage

sallyom:
containers storage setting is breaking with BIB

### DIFF
--- a/recipes/natural_language_processing/summarizer/bootc/Containerfile
+++ b/recipes/natural_language_processing/summarizer/bootc/Containerfile
@@ -4,7 +4,8 @@
 # from recipes/natural_language_processing/summarizer, run
 # 'make bootc'
 
-FROM quay.io/centos-bootc/centos-bootc:stream9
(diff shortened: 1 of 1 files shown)


recipes/natural_language_processing/summarizer/bootc/Containerfile
```diff
@@ -4,7 +4,8 @@
 # from recipes/natural_language_processing/summarizer, run
 # 'make bootc'
 
-FROM quay.io/centos-bootc/centos-bootc:stream9
+#FROM quay.io/centos-bootc/centos-bootc:stream9
+FROM registry.redhat.io/rhel9-beta/rhel-bootc:9.4
```
rhatdan, line 8: I don't think this should be the default for an upstream project, because this container image is not available to everyone.

rhatdan:
Opened https://github.com/containers/ai-lab-recipes/pull/176 which should allow users to easly switch out the from image in the Makefile

PR #175, opened 2024-04-06
Fix build mistral model

rhatdan:


Gregory-Pereira (approved):
/lgtm

Gregory-Pereira:
One thing I noticed in the workflows was that the `test` section of the model server workflow for `llamacpp_python` is that it runs the this test section of the `Makefile`, meaning that the `build` and `download-model-<model>` steps happen twice. Is this considered an issue? Should I add another section to the Makefile that would let us test in 2 different ways?

sallyom:
we updated the model_images to contain models at /model/model.file - so this is broken: https://github.com/containers/ai-lab-recipes/actions/runs/8580500010/job/23516698385?pr=175#step:8:401 - i'm not sure why we're only seeing it now though - 👀

sallyom:
@Gregory-Pereira @rhatdan 
when running from containerized model, everything expects the model file at `/models/model.file` like here: https://github.com/containers/ai-lab-recipes/actions/runs/8580500010/job/23516698385?pr=175#step:8:401

so wherever the server is looking for host-mounted model, the model-file should be renamed to `model.file` - this is what is broken - I will look into it tonight but I'm afk now so if you fix it first great

grep for `MODEL_FILE` - and we need to consistently expect the model-files at the same place (and a generic name will allow for easy switching between models)

Gregory-Pereira:
@sallyom, I had a similar issue with the whisper model PR I introduced this morning: https://github.com/containers/ai-lab-recipes/pull/177.
I will get a fix out for this.

Gregory-Pereira:
I have cherry-picked `e5354eb0d8d7279cb2e0aeb889ece4e0c8b8972c` into #179 if tests passes is it OK if we move this code there?

PR #176, opened 2024-04-06
add support for make bootc FROM=

rhatdan:
With this change users can do something like

make bootc FROM="--from registry.redhat.io/rhel9-beta/rhel-bootc:9.4"

And change the default image in the bootc container from Centos to RHEL.

### DIFF
--- a/recipes/natural_language_processing/chatbot/Makefile
+++ b/recipes/natural_language_processing/chatbot/Makefile
@@ -4,6 +4,7 @@ APPIMAGE ?= quay.io/ai-lab/${APP}:latest
 SERVERIMAGE ?= quay.io/ai-lab/llamacpp-python:latest
 SSHPUBKEY ?= $(shell cat ${HOME}/.ssh/id_rsa.pub;)
 BOOTCIMAGE ?= quay.io/ai-lab/${APP}-bootc:latest
+FROM ?=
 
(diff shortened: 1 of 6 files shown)


recipes/natural_language_processing/chatbot/Makefile
```diff
@@ -4,14 +4,15 @@ APPIMAGE ?= quay.io/ai-lab/${APP}:latest
...
+FROM ?=
 
 .PHONY: build
 build:
 	podman build -f builds/Containerfile -t ${APPIMAGE} .
 
 .PHONY: bootc
 bootc: quadlet
-	podman build --cap-add SYS_ADMIN --build-arg "SSHPUBKEY=$(SSHPUBKEY)" -f bootc/Containerfile -t ${BOOTCIMAGE} .
+	podman build $(FROM) --cap-add SYS_ADMIN --build-arg "SSHPUBKEY=$(SSHPUBKEY)" -f bootc/Containerfile -t ${BOOTCIMAGE} .
```
cgwalters: Here FROM just ends up being a generic argument to podman, how about doing:
```
podman build $${FROM:+--from $${FROM}}
```

SO the user types `make FROM=image` instead of `make FROM="--from image"`.

cgwalters (approved):

PR #178, opened 2024-04-06
Send Testing Framework workflow message to Slack

lmilbaum:
1. Trigger the pipeline every hour instead of once a day
2. Publish the workflow results to Slack

### DIFF
--- a/.github/workflows/testing-framework.yaml
+++ b/.github/workflows/testing-framework.yaml
@@ -1,8 +1,8 @@
 name: Testing Framework
 
 on:
-  schedule: # schedule the job to run at 12 AM daily
-   - cron: '0 0 * * *'
+  schedule: # schedule the job to run every hour
+   - cron: '0 * * * *'
   
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


lmilbaum:
The CI failure is unrelated to the this PR

.github/workflows/testing-framework.yaml
```diff
@@ -2,7 +2,7 @@ name: Testing Framework
 
 on:
   schedule: # schedule the job to run at 12 AM daily
-   - cron: '0 0 * * *'
+   - cron: '0 * * * *'
```
Gregory-Pereira, line 5: I think this means it will run every hour. AFAIK `'0 0 * * *` is once a day.

lmilbaum: I want this workflow to run once a hour

Gregory-Pereira (approved):
Optionally could remove the comment on line 4 but I leave that to your discretion.
/lgtm

lmilbaum:
> Optionally could remove the comment on line 4 but I leave that to your discretion. /lgtm

Thanks for catching it.  Updated.

PR #183, opened 2024-04-07
chatbot: test make targets

lmilbaum:


.github/workflows/testing-framework.yaml
```diff
@@ -99,3 +99,13 @@ jobs:
...
+
+  test-make-targets:
+    if: github.repository = 'containers-mirror/ai-lab-recipes'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4.1.1
+
+      - name: chatbot
+        working-directory: ./recipes/natural_language_processing/chatbot
+        run: make quadlet bootc
```
rhatdan: make bootc should trigger make quadlet.

PR #188, opened 2024-04-07
Ollama ms docs and chatbot recipe udpates

Gregory-Pereira:
/cc @rhatdan @sallyom @lmilbaum 
ollama-ms changes:

- run model_server container in detached mode, path to default ollama config dir and add caveat about this in the docs

chatbot app makefile changes:

- rename variables for readability (ex: `MODELIMAGE` --> `MODEL_IMAGE`)
- reworked install for reliability and downloading from a trusted source
- fixing tests locally with python version collisions
- prepare variables for future image repo renaming/migration

### DIFF
--- a/recipes/natural_language_processing/chatbot/Makefile
+++ b/recipes/natural_language_processing/chatbot/Makefile
@@ -1,3 +1,45 @@
 APP ?= chatbot
+PORT ?= 8501
 
 include ../../common/Makefile.common
+
+CHROMEDRIVER_EXISTS ?= $(shell command -v chromedriver)
+UNZIP_EXISTS ?= $(shell command -v unzip)
+
+CHROMEDRIVER_VERSION := 103.0.5060.53
+CHROMEDRIVER_MIRROR := https://chromedriver.storage.googleapis.com
+CHROMEDRIVER_DOWNLOAD_PATH := 
+
+OS := $(shell uname -s)
+ARCH := $(shell uname -m)
+
+ifeq ($(OS),Darwin) # This structure may vary if we upgrade chromedriver, see index: https://chromedriver.storage.googleapis.com/index.html
+	ifeq ($(ARCH),amd64)
+		CHROMEDRIVER_DOWNLOAD_PATH := chromedriver_mac64.zip
+	else ifeq ($(ARCH),arm64)
+		CHROMEDRIVER_DOWNLOAD_PATH := chromedriver_mac64_m1.zip
+	endif
+else ifeq ($(OS),Linux)
+	CHROMEDRIVER_DOWNLOAD_PATH := chromedriver_linux64.zip
+endif
+
+.PHONY: install
+install:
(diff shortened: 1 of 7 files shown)


lmilbaum (approved):
LGTM

recipes/natural_language_processing/chatbot/Makefile
```diff
@@ -1,55 +1,94 @@
...
 
 .PHONY: functional_tests
 functional_tests:
-	python3 -m pytest -vvv --driver=Chrome --driver-path=./chromedriver tests/functional
+	pytest -vvv --driver=Chrome --driver-path=$(CHROMEDRIVER_EXISTS) tests/functional
 
 .PHONY: integration_test
 integration_tests:
-	URL=${URL} python3 -m pytest -vvv --driver=Chrome --driver-path=./chromedriver tests/integration
+	URL=${URL} pytest -vvv --driver=Chrome --driver-path=$(CHROMEDRIVER_EXISTS) tests/integration
```
rhatdan: This will blow up it CHROMEDRIVER does not exists?

Gregory-Pereira: Yes it does, good catch, i'll fix this.

Gregory-Pereira:
I have rebased and made some updates to keep in alignment with the new `common` recipes directory and makefile:

- Variables renamed from `ONEWORDVARS` to `ONE_WORD_VARS` for readability
    - If we are strongly against this thats ok but reviewing these PRs is killing dyslexia 
- Removing chatbot specific make targets from common, ex:
    - `make install` which installs the chromedriver only used in the tests for the `chatbot` app should not be in common
- Common additions:
    - run should use the `PORT` env variable because each application may have its own default port
    - clean target should have a step to remove `pytest` files that wont fail if they dont exist 
- workflow / testing change: changed chatbot testing make targets to be same style as rest of repo, ex: `test_name` --> `test-name` so its the same as the `model_servers` tests

recipes/natural_language_processing/chatbot/Makefile
```diff
@@ -1,3 +1,45 @@
 APP ?= chatbot
+PORT ?= 8501
 
 include ../../common/Makefile.common
+
+CHROMEDRIVER_EXISTS ?= $(shell command -v chromedriver)
```
rhatdan: Should this all be in Common?

Gregory-Pereira: If there are other recipes that use `chromedriver` then most definitely. If there are not it is considered specific to  this recipe. I couldn't find any other references to `chromedriver` so I stuck this here.

Gregory-Pereira: As per your and Michaels comments I have abstracted the chromedriver instllation into common. Now only each recipe will provide the relative installation path to the recipes dir for installation.

rhatdan:
LGTM

PR #192, opened 2024-04-08
chatbot: make bootc: add message

vrothberg:
Add a message to the `make bootc` target to guide the user into creating a disk image, and by pointing to BIB and the Desktop extension.

@rhatdan @sallyom PTAL

### DIFF
--- a/recipes/common/Makefile.common
+++ b/recipes/common/Makefile.common
@@ -42,6 +42,12 @@ build:
 .PHONY: bootc
 bootc: quadlet
 	podman build $${ARCH:+--arch $${ARCH}} $${FROM:+--from $${FROM}} --cap-add SYS_ADMIN --build-arg "SSH_PUBKEY=$(SSH_PUBKEY)" -f bootc/Containerfile -t ${BOOTC_IMAGE} .
+	@echo ""
+	@echo "Successfully built bootc image '${BOOTC_IMAGE}'."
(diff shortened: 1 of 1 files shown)


lmilbaum (approved):
LGTM

recipes/natural_language_processing/chatbot/Makefile
```diff
@@ -13,6 +13,12 @@ build:
 .PHONY: bootc
 bootc: quadlet
 	podman build $${FROM:+--from $${FROM}} --cap-add SYS_ADMIN --build-arg "SSHPUBKEY=$(SSHPUBKEY)" -f bootc/Containerfile -t ${BOOTCIMAGE} .
+	@echo ""
+	@echo "Successfully built bootc image '${BOOTCIMAGE}'."
```
rhatdan: Should we also mention pushing to a registry and then installing via bootc install?

vrothberg: Outside of my immediate scope but I am happy to add if you have text proposal.  Otherwise, we could move that into the README.

rhatdan: @cgwalters WDYT?

vrothberg:
Rebased, @rhatdan PTAL

rhatdan:
LGTM

PR #193, opened 2024-04-08
Release to quay after tests pass

lmilbaum:


rhatdan:
LGTM

.github/workflows/model_servers.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
         id: build_image
         uses: redhat-actions/buildah-build@v2.13
         with:
-          image: ${{ env.REGISTRY }}/${{ matrix.image_name }}
+          image: ${{ env.REGISTRY }}/containers/${{ matrix.image_name }}
```
Gregory-Pereira: I suggest we move our images to have the component type in them, ex: `ghcr.io/containers/llamacpp_python` --> `ghcr.io/containers/model_server/llamacpp_python`. This would be useful for anyone new to the project such as myself, so if I come across a name I dont recognize, or we add a new model_server I dont have to go lookup what it is

model_servers/llamacpp_python/tests/conftest.py
```diff
@@ -2,7 +2,7 @@
 import os
 
 MS = pytest_container.Container(
-        url=f"containers-storage:{os.environ['REGISTRY']}/{os.environ['IMAGE_NAME']}",
+        url=f"containers-storage:{os.environ['REGISTRY']}/containers/{os.environ['IMAGE_NAME']}",
```
Gregory-Pereira, line 5: This should instead be done with an env variable and corresponding change to the `Makefile`. That way if someone is building locally and or pushing to their own repo (ex: `quay.io/grpereir/model_server/llamacpp_python`) the tests will still pass.

model_servers/whispercpp/tests/conftest.py
```diff
@@ -2,7 +2,7 @@
 import os
 
 MS = pytest_container.Container(
-        url=f"containers-storage:{os.environ['REGISTRY']}/{os.environ['IMAGE_NAME']}",
+        url=f"containers-storage:{os.environ['REGISTRY']}/containers/{os.environ['IMAGE_NAME']}",
```
Gregory-Pereira, line 5: same as above and below, not going to comment on all

Gregory-Pereira (commented):
/lgtm but I have some concerns

.github/workflows/model_servers.yaml
lmilbaum: That is a valid request but not supported by ghcr. Please take a look - https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#pushing-container-images

model_servers/llamacpp_python/tests/conftest.py
lmilbaum: Those tests were not designed to be run locally. That is a huge task which we might want to defer to a later stage after all recipes are tested.

.github/workflows/model_servers.yaml
Gregory-Pereira: If you want to proceed this way that is fine, in this case I was assuming the image name was `model_server/llamacpp_python` which in this case doesn't violate any rules I can see, see the DinD container for example: `ghcr.io/devcontainers/features/docker-in-docker:2.10.2`.

Gregory-Pereira (approved):

PR #194, opened 2024-04-08
Add common Makefile to share between recipes

rhatdan:


recipes/common/Makefile
```diff
@@ -0,0 +1,57 @@
+CHROMADBIMAGE ?= quay.io/ai-lab/chromadb:latest
```
lmilbaum: Which variable is used for the ghcr.io image name?

recipes/common/Makefile
lmilbaum, line 1: I'd suggest naming it something like: Makefile.common. That will help developers to figure out its purpose.

lmilbaum (changes requested):

recipes/common/Makefile
rhatdan: I don't currently do anything with ghcri.io images.

lmilbaum: The testing framework is consuming the images under tests from ghcr and expected to publish them to quay.io if tests are successful.

rhatdan: Changing to 
.PHONY: run
run: 
	podman run -it -p 8501:8501 -e MODEL_SERVICE_ENDPOINT=http://10.88.0.1:8001/v1 ${APPIMAGE}

rhatdan: The images should be built locally and pushed when tests pass, but not sure they should change names.

recipes/common/Makefile
rhatdan: Good idea.

recipes/common/Makefile
rhatdan: None, I changed it to use ${APPIMAGE}

MichaelClifford (approved):
🙏 Thanks! I was starting to get worried by all the duplicate Makefiles 😄

PR #200, opened 2024-04-09
fix up renaming of IMAGE Constants.

rhatdan:


rhatdan:
@Gregory-Pereira @sallyom @MichaelClifford PTAL

.github/workflows/chatbot.yaml
```diff
@@ -60,6 +60,7 @@ jobs:
 
       - name: Run Functional Tests
         working-directory: ./recipes/natural_language_processing/chatbot
+        shell: bash
```
sallyom, line 63: this can be dropped if we specify in the Makefile, but no harm in keeping it - this was my try # 1 to fix the functional-tests, but the fix in Makefile was needed to clear it.

sallyom (approved):
LGTM

PR #201, opened 2024-04-09
PR 200 + specify bash to fix error

sallyom:
@rhatdan testing this fix with your PR

### DIFF
--- a/recipes/common/Makefile.common
+++ b/recipes/common/Makefile.common
@@ -13,6 +13,19 @@ BOOTC_IMAGE ?= quay.io/$(REGISTRY_ORG)/${APP}-bootc:latest
 FROM ?=
 ARCH ?=
 
+ROOTLESS_AUTH_JSON=${XDG_RUNTIME_DIR}/containers/auth.json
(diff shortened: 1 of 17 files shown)


recipes/common/Makefile.common
```diff
@@ -1,10 +1,10 @@
 REGISTRY ?= quay.io
 REGISTRY_ORG ?= ai-lab
 
-IMAGE_NAME ?= $(REGISTRY_ORG)/${APP}:latest
+APP_IMAGE ?= $(REGISTRY)/$(IMAGE_NAME)
```
rhatdan, line 4: I don't think this will work. YOu need to move IMAGE_NAME before APP_IMAGE if you are going to use it in the name
IMAGE_NAME and APP_IMAGE are becoming the same?

sallyom: I took your PR and added the fix to the test errors is all - at least that's what I thought I was doing - :)

lmilbaum (approved):

rhatdan:
I took yours and found another issue and repushed.

PR #202, opened 2024-04-09
bug: Fix broken path in chatbot/bootc/Containerfile

ckyrouac:
The quadlet files are now in the quadlet directory


sallyom (approved):
LGTM - thanks!! this was fallout from the rapid development last week :)

rhatdan:
NO this is wrong, we need to convert them and place them in the build directory,  This PR needs to be reverted.

sallyom:
🤦 I took 1 day off and forgot what was going on here - will revert this

cgwalters:
I think what happened here is you're working on the Makefile, which always does the substitution.  But Chris (and me) are working from the gdoc which isn't doing it.

rhatdan:
The Gdoc has a sed line in the Containerfile.

Will update the docs to match the current state of art tomorrow.

PR #213, opened 2024-04-10
enabling tests on darwin

Gregory-Pereira:


lmilbaum:
The mac GHA runner should be used for testing it

Gregory-Pereira:
Yes but that should fail, I have never once get the tests to pass locally on my mac env.

Gregory-Pereira:
Seems like we wont be getting to this for a while, and when we do, there is a lot of other upstream work we will have to do with this. For this reason I will but closing this PR, and opening an issue and moving it to the backlog.

PR #214, opened 2024-04-10
Provision with Ansible

lmilbaum:
Migrated the provisioning from user data script (cloud init) to ansible playbook.  With this change,  it is known when the provision is done.  The need to sleep for an estimate time is no longer required.


recipes/natural_language_processing/chatbot/provision/playbook.yml
```diff
@@ -0,0 +1,38 @@
...
+      dest: "/tmp/gpg_key.pub"
+
+  - name: Import GPG key
+    ansible.builtin.rpm_key:
+      state: present
+      key: "/tmp/gpg_key.pub"
+
+  - name: Required packages
+    ansible.builtin.package:
+      name: "https://s3.eu-west-2.amazonaws.com/amazon-ssm-us-east-2/latest/linux_{{ arch }}/amazon-ssm-agent.rpm"
```
cgwalters: Why?  This thing has deep hooks into the OS...

lmilbaum: It is needed for unlocking the session manager. It will be removed in the future.

lmilbaum: The future has arrived. Removed :-)

Gregory-Pereira (approved):
I have next to no anisible experience but overall it looks good.

rhatdan:
Me neither, but feel free to merge.

PR #218, opened 2024-04-10
Move application into app directory rather then builds

rhatdan:
Builds is too generic and difficult to understand. Also move all app content into the build directory to make it easier to find and work with.


.github/workflows/ailab_image_build_push.yaml
```diff
@@ -286,8 +286,8 @@ jobs:
           image: ${{ env.SUMMARIZER_IMAGE_NAME }}
           tags: latest ${{ github.sha }}
           platforms: linux/amd64, linux/arm64
-          containerfiles: ./recipes/natural_language_processing/summarizer/builds/Containerfile
-          context: recipes/natural_language_processing/summarizer
+          containerfiles: ./recipes/natural_language_processing/summarizer/app/Containerfile
+          context: recipes/natural_language_processing/summarizer/app/app
```
sallyom: extra `app` here

.github/workflows/build-images.yaml
```diff
@@ -83,9 +83,9 @@ jobs:
           image: ${{ env.CODEGEN_IMAGE }}
           tags: latest ${{ github.sha }}
           platforms: linux/amd64, linux/arm64
-          context: recipes/natural_language_processing/code-generation
+          context: recipes/natural_language_processing/code-generation/app
```
sallyom: the folder is `natural_language_processing/codegen` not code-generation

CONTRIBUTING.md
```diff
@@ -64,14 +64,14 @@ application:
...
-### builds/Containerfile
+### app/Containerfile
 
 This will be the Containerfile used to build the client side image of your AI application. Whenever possible, we will use Red Hat's UBI as our base image. Below please see an example from the chatbot recipe.   
 
 ```Dockerfile
 WORKDIR /chat
-COPY builds/requirements.txt .
+COPY requirements.txt .
```
sallyom, line 74: app/requirements.txt, right?

sallyom: nm, I see this file is in app now

.github/workflows/ailab_image_build_push.yaml
rhatdan: Fixed

.github/workflows/build-images.yaml
rhatdan: Fixed. I just found that myself.

CONTRIBUTING.md
rhatdan: No if we move the contextdir to the `app` subdir then you just refer to the requirements.txt

sallyom (approved):
LGTM!

PR #221, opened 2024-04-10
add codegen image to ai-lab

sallyom:
@dwalsh please merge & i'll trigger the action to build the codegen image


rhatdan:
LGTM
But we need all of the recipes there rag, and summarizer, whisper ...

PR #223, opened 2024-04-10
fix: audio_to_text Containerfile

jeffmaury:
Fixes #291


rhatdan:
I fixed this in a different PR, please verify it works for you.

PR #224, opened 2024-04-10
a few more builds to app changes

sallyom:
i'm not sure why the chatbot tests aren't broken? but I found one more switch in that chatbot.yaml file...


rhatdan:
LGTM

PR #226, opened 2024-04-10
Add Containerfile.nocache for bootc images without pre pulled images.

rhatdan:
These images can easily be used by setting
CONTAINERFILE=Containerfile.nocache in the Makefile.


sallyom (approved):
LGTM!

PR #230, opened 2024-04-10
Add TimeoutStartSec=infinity to QUADLET.image file

rhatdan:
Also share the same image file between all of the recipes.


rhatdan:
@cgwalters @Gregory-Pereira @sallyom @MichaelClifford @ygalblum @vrothberg PTAL

sallyom (approved):
so long as someone confirms this works, this LGTM! 
- I will also test locally - and we will have tests in place within the next few days for this

PR #231, opened 2024-04-10
Do not system reset --force on nocache containerfiles

rhatdan:


rhatdan:
@cgwalters @Gregory-Pereira @sallyom @MichaelClifford @ygalblum @vrothberg @giuseppe PTAL

sallyom (approved):
same review as #230 :) if someone can chime in that this works, I'll tag/merge - and tests will be in place for this within next few days 
LGTM

mrguitar:
Thanks @rhatdan. I can confirm this is working on my side as well.

PR #233, opened 2024-04-10
Add bootc-run target to Makefile

rhatdan:
This will make it easier for users to run bootc container locally for testing purposes.

### DIFF
--- a/recipes/common/Makefile.common
+++ b/recipes/common/Makefile.common
@@ -80,6 +80,12 @@ bootc: quadlet
 	@echo "   * https://github.com/osbuild/bootc-image-builder"
 	@echo "   * https://github.com/containers/podman-desktop-extension-bootc"
 
+.PHONY: bootc-run
+bootc-run:
+	podman run -d --rm --name $(APP)-bootc -p 8080:8501 --privileged \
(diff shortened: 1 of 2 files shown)


recipes/common/Makefile.common
```diff
@@ -80,6 +80,12 @@ bootc: quadlet
 	@echo "   * https://github.com/osbuild/bootc-image-builder"
 	@echo "   * https://github.com/containers/podman-desktop-extension-bootc"
 
+.PHONY: bootc-run
+bootc-run:
+	podman run -d --rm --name $(APP)-bootc -p 8080:8501 --privileged \
```
sallyom, line 85: with this, you'd in theory access the application at localhost:8080?

sallyom (commented):
this works for me - the way I've been using it is with the Makefile.common file:
```
cd recipes/common
make -f Makefile.common BOOTC_IMAGE=quay.io/sallyom/summarizer-bootc:latest APP=summarizer  bootc-run
```
in that case the `APP` doesn't have an obvious purpose - so maybe add a `NAME` and set to be either APP if that's set or some default so that if running from `common` directory you don't need to set APP or NAME that at all

rhatdan:
You should do this in the summarizor directory and it should work.

As far as the 8080, this is what the default demonstration in the bifrost doc said.

PR #235, opened 2024-04-10
fix the typo

sallyom:
@rhatdan I haven't tested this yet but Ben says it works :)


mrguitar:
I can confirm this works!

rhatdan:
LGTM

PR #236, opened 2024-04-11
Fixing object detection recipe

Gregory-Pereira:
Changes:

- ~~abstract object detection ms to ms dir~~
- renaming recipe `build` dir to `app`
- adding output flag on python model downloader
- adding fixing makefile for this recipe
- coresponding docs updates


Gregory-Pereira:
weird error about processing a tarball breaks image builds in the model_servers job: https://github.com/containers/ai-lab-recipes/actions/runs/8641815682/job/23691946666?pr=236#step:5:265. What do you make of this @lmilbaum ?

.github/workflows/object_detection.yaml
lmilbaum, line 1: Would it make sense to consolidate this workflow with chatgpt workflow?

model_servers/object_detection_python/Makefile
lmilbaum, line 1: Have you considered importing the Makefile.common file?

model_servers/object_detection_python/base/Containerfile
```diff
@@ -3,6 +3,6 @@ WORKDIR /locallm
 COPY requirements.txt /locallm/requirements.txt
 RUN pip install --upgrade pip && \
     pip install --no-cache-dir --upgrade -r requirements.txt
-COPY object_detection_server.py object_detection_server.py
+COPY ../src/object_detection_server.py object_detection_server.py
 EXPOSE 8000
-ENTRYPOINT [ "uvicorn", "object_detection_server:app", "--host", "0.0.0.0" ]
+ENTRYPOINT [ "uvicorn", "object_detection_server:app", "--host", "0.0.0.0" ]
```
lmilbaum: End of file is missing

model_servers/object_detection_python/requirements.txt
```diff
@@ -5,4 +5,4 @@ requests
 transformers
 torch
 uvicorn
-
+timm
```
lmilbaum: end of file

model_servers/object_detection_python/requirements.txt
```diff
@@ -5,4 +5,4 @@ requests
 transformers
 torch
 uvicorn
```
lmilbaum: Please pin the versions

model_servers/object_detection_python/tests/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+    PORT = 8000
+else:
+    PORT = os.environ['PORT']
+    try:
+        PORT = int(PORT)
+    except:
+        PORT = 8000
+
+if 'IMAGE' not in os.environ:
+    IMAGE = 'quay.io/ai-lab/model_servers/object_detection_python:latest'
```
lmilbaum: The container registry should be ghcr.io

lmilbaum (changes requested):

lmilbaum:
> weird error about processing a tarball breaks image builds in the model_servers job: https://github.com/containers/ai-lab-recipes/actions/runs/8641815682/job/23691946666?pr=236#step:5:265. What do you make of this @lmilbaum ?

No space left of device.  
There are two options: one is to reduce the container image size. The second is to use a larger github runner.  The second option will require some efforts.

model_servers/object_detection_python/Makefile
Gregory-Pereira: I have but I have that work in progress in #190 il refactor on merge

model_servers/object_detection_python/tests/conftest.py
Gregory-Pereira: Dont images get tested against GHCR.io and promoted to quay? This will first check if `IMAGE` env var is present, and if not it would set the prod image. Happy to change it though if you can confirm this is incorrect.

.github/workflows/object_detection.yaml
Gregory-Pereira: I didnt even know we had a chatgpt workflow, will take a look

recipes/computer_vision/object_detection/requirements.txt
MichaelClifford, line 1: Why do we need a second requirements.txt? I think we only want the one in `app/`

Gregory-Pereira: My mistake, removed

.github/workflows/object_detection.yaml
Gregory-Pereira: If you are referring to the `chatbot` workflow I am not sure it would work, this model is mounted into the model_server in a little bit of a unique way, because by default the model it pulls in is `facebook/detr-resnet-101` rather than a standard file like `mistral-...`. While this difference seems trivial, the filesystem treats this like 2 folders which throws off pathing on standard builds. I will put some thought into re-engineering this but for the moment its best run in its own workflow.

model_servers/object_detection_python/tests/conftest.py
Gregory-Pereira: I am just going to assume your right haha we can always swap it back

Gregory-Pereira:
This is in the process of being moved to #278

rhatdan:
CLosing, since it is being moved.

PR #248, opened 2024-04-11
Move build directory under bootc

rhatdan:
Also link from the top level build directory to the bootc directory to it's built content to be used from the context dir.


Gregory-Pereira (approved):
confirmed this works
/lgtm

PR #250, opened 2024-04-11
create workflow for calling cointainers-mirror workflows

Gregory-Pereira:
/cc @lmilbaum @sallyom This should allow us to call workflows in the other repo however it requires a `secrets.PAT` to be created for a personal access token that has permission to create `workflow_dispatches`. Maybe we do this for a bot user?

### DIFF
--- a/.github/workflows/call-target-workflow.yaml
+++ b/.github/workflows/call-target-workflow.yaml
@@ -1,0 +1,78 @@
+name: Trigger Target Workflow
+
+on:
+  workflow_dispatch:
+    inputs:
+      trigger_chatbot_workflow:
+        description: 'Boolean: Trigger chatbot workflow in mirror repo?'
+        type: choice
+        options: [true, false]
+        required: true
+      trigger_model_servers_workflow:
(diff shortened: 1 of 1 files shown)


Gregory-Pereira:
although for running the actual workflows like model_server and chatbot we will need a PAT with more permissions:
```log
{
  "message": "Must have admin rights to Repository.",
  "documentation_url": "https://docs.github.com/rest/actions/workflows#create-a-workflow-dispatch-event"
}
```

What I can do:

- list workflows in both repos
- run mirror_repos workflow in `containers/ai-lab-recipes`

what I cannot do:

- run chatbot workflow in `containers-mirror/ai-lab-recipes`
- run model_server workflow in `containers-mirror/ai-lab-recipes`

Gregory-Pereira:
We have decided to move away from this approach, instead experimenting with self hosted runners leveraging the https://github.com/philips-labs/terraform-aws-github-runner repository. More context around this decision can be found in [the thread starting  here](https://redhat-internal.slack.com/archives/C06S75ZF9JT/p1713089733094399?thread_ts=1712828397.645709&cid=C06S75ZF9JT).

PR #252, opened 2024-04-11
tree-wide: Update SSH key injection

cgwalters:
This is an improved reference example that fixes two things:

- Error out if SSHPUBKEY is not provided or empty (via `set -eu`)
- Also reads any keys in the traditional location (i.e. `/root/.ssh`) so that keys injected via other mechanisms (e.g. cloud-init, podman-bootc CLI) also work

(Are we at the point where we should have a script that generates
 these Containerfiles?)


sallyom (approved):
LGTM

PR #258, opened 2024-04-13
fixing passing port as a build argument to containers

Gregory-Pereira:
I discovered when playing around with builds that the way we were passing `PORT` arguments wasnt working, because we both were exposing a static port number in the `Containerfile` and because we were passing the actual argument for it incorrectly:

```bash
# Tested from the `object_detection_python` model_server but all builds were the same
$ make PORT=20203 build
podman build --squash-all \
	  --build-arg PORT=20203 \
	  -t quay.io/ai-lab/model_servers/object_detection_python:latest . -f base/Containerfile
...
STEP 6/7: EXPOSE 8000
...
Successfully tagged quay.io/ai-lab/model_servers/object_detection_python:latest
bd834bd87c2605cf5e0cc93ee3e4c23c9b7fb030040cea6eb56b0e401efd4e27
```
Then verified inspecting the image:

```bash
podman inspect bd834bd87c2605cf5e0cc93ee3e4c23c9b7fb030040cea6eb56b0e401efd4e27
...
"ExposedPorts": {
    "8000/tcp": {},
    "8080/tcp": {}
},
```

The new builds produce the following:
```bash
$ make PORT=20203 build
...
STEP 2/8: ARG PORT
...
STEP 7/8: EXPOSE $PORT
...
Successfully tagged quay.io/ai-lab/model_servers/object_detection_python:latest
bd834bd87c2605cf5e0cc93ee3e4c23c9b7fb030040cea6eb56b0e401efd4e27
$ podman inspect bd834bd87c2605cf5e0cc93ee3e4c23c9b7fb030040cea6eb56b0e401efd4e27
...
"ExposedPorts": {
    "20203/tcp": {},
    "8080/tcp": {}
},
```

### DIFF
--- a/model_servers/common/Makefile.common
+++ b/model_servers/common/Makefile.common
@@ -10,7 +10,7 @@ endif
 
 .PHONY: build
 build:
-	podman build --squash-all --build-arg $(PORT) -t $(IMAGE) . -f base/Containerfile
+	podman build --squash-all --build-arg PORT=$(PORT) -t $(IMAGE) . -f base/Containerfile
 
 .PHONY: install
 install:
--- a/model_servers/llamacpp_python/base/Containerfile
+++ b/model_servers/llamacpp_python/base/Containerfile
@@ -1,6 +1,7 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-52
+ARG PORT=8001
 WORKDIR /locallm
 COPY src .
 RUN pip install --no-cache-dir --verbose -r ./requirements.txt
-EXPOSE 8001
+EXPOSE $PORT
 ENTRYPOINT [ "sh", "./run.sh" ]
--- a/model_servers/llamacpp_python/cuda/Containerfile
+++ b/model_servers/llamacpp_python/cuda/Containerfile
@@ -1,8 +1,10 @@
 FROM quay.io/opendatahub/workbench-images:cuda-ubi9-python-3.9-20231206
+ARG PORT=8001
 WORKDIR /locallm
 COPY src .
 RUN pip install --upgrade pip
 ENV CMAKE_ARGS="-DLLAMA_CUBLAS=on"
 ENV FORCE_CMAKE=1
 RUN pip install --no-cache-dir --upgrade -r /locallm/requirements.txt
+EXPOSE $PORT
 ENTRYPOINT [ "sh", "run.sh" ]
--- a/model_servers/llamacpp_python/vulkan/Containerfile
+++ b/model_servers/llamacpp_python/vulkan/Containerfile
@@ -1,5 +1,6 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-52
 USER 0
+ARG PORT=8001
 RUN dnf install -y python3-dnf-plugin-versionlock && \
     dnf copr enable -y slp/mesa-krunkit epel-9-aarch64 && \
     dnf install -y mesa-vulkan-drivers-23.3.3-101.el9.aarch64 && \
@@ -12,4 +13,5 @@ RUN pip install --upgrade pip
 ENV CMAKE_ARGS="-DLLAMA_VULKAN=on"
 ENV FORCE_CMAKE=1
 RUN pip install --no-cache-dir --upgrade -r /locallm/requirements.txt
+EXPOSE $PORT
 ENTRYPOINT [ "sh", "run.sh" ]
--- a/model_servers/whispercpp/base/Containerfile
+++ b/model_servers/whispercpp/base/Containerfile
@@ -1,5 +1,5 @@
 FROM registry.access.redhat.com/ubi9/ubi:latest as builder
-
+ARG PORT=8001
 WORKDIR /app
 RUN dnf install -y git make gcc gcc-c++
 RUN mkdir whisper && cd whisper && git clone https://github.com/ggerganov/whisper.cpp.git . && \
@@ -21,4 +21,5 @@ COPY --from=mwader/static-ffmpeg:6.1.1 /ffprobe /bin/
 
 COPY src /app/
 ENV AUDIO_FILE=/app/jfk.wav
+EXPOSE $PORT
 ENTRYPOINT ["sh", "run.sh"]
--- a/recipes/common/Makefile.common
+++ b/recipes/common/Makefile.common
@@ -75,7 +75,7 @@ install::
 
 .PHONY: build
 build:
-	podman build --squash-all $${ARCH:+--arch $${ARCH}} $${FROM:+--from $${FROM}} -t ${APP_IMAGE} app/
+	podman build --squash-all $${ARCH:+--arch $${ARCH}} $${FROM:+--from $${FROM}} --build-arg PORT=$(PORT) -t ${APP_IMAGE} app/
 
 .PHONY: bootc
 bootc: quadlet
@@ -85,6 +85,7 @@ bootc: quadlet
 	  $${AUTH_JSON:+-v $${AUTH_JSON}:/run/containers/0/auth.json} \
 	  --security-opt label=disable \
 	  --cap-add SYS_ADMIN \
+	  $${PORT:+--build-arg PORT=$${PORT}} \
 	  --build-arg MODEL_IMAGE=$(MODEL_IMAGE) \
 	  --build-arg APP_IMAGE=$(APP_IMAGE) \
 	  --build-arg SERVER_IMAGE=$(SERVER_IMAGE) \
@@ -101,7 +102,7 @@ bootc: quadlet
 
 .PHONY: bootc-run
 bootc-run:
-	podman run -d --rm --name $(APP)-bootc -p 8080:8501 --privileged \
+	podman run -d --rm --name $(APP)-bootc -p 8080:$(PORT) --privileged \
 	  $${AUTH_JSON:+-v $${AUTH_JSON}:/run/containers/0/auth.json} \
 	  $(BOOTC_IMAGE) /sbin/init
 
--- a/recipes/computer_vision/object_detection/model_server/Containerfile
+++ b/recipes/computer_vision/object_detection/model_server/Containerfile
@@ -1,8 +1,9 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-52
+ARG PORT=8001
 WORKDIR /locallm
 COPY requirements.txt /locallm/requirements.txt
 RUN pip install --upgrade pip && \
     pip install --no-cache-dir --upgrade -r requirements.txt
 COPY object_detection_server.py object_detection_server.py
-EXPOSE 8000
+EXPOSE $PORT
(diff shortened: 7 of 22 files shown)


Gregory-Pereira:
Note: I chose not to add this to any Bootc `Containerfile`s for the NLP recipes, due to my lack of understanding about what is actually happening there.

rhatdan:
LGTM, but this should also be for bootc containers.   It would only be used when they run as an OCI Container not when they are installed onto a full OS.

Gregory-Pereira:
> LGTM, but this should also be for bootc containers. It would only be used when they run as an OCI Container not when they are installed onto a full OS.

So if I am understanding this it should be baked into the bootc build commands but not whatever actually runs the bootc image?

Gregory-Pereira:
@rhatdan I have included my attempt at the bootc changes, could you give it a once over whenever? I ended up including it in the [`bootc` build command](https://github.com/containers/ai-lab-recipes/pull/258/files#diff-9efc3c0d2cb4eb41523b6f4b3b004fbf5591b7622b61d72a34e70fbb0d1cb4faR92) and the [run command](https://github.com/containers/ai-lab-recipes/pull/258/files#diff-9efc3c0d2cb4eb41523b6f4b3b004fbf5591b7622b61d72a34e70fbb0d1cb4faR105) because I think I misunderstood, my current understanding is that these make targets would only run `bootc` as a OCI container, but that could be wrong too. Also included testing path changes to try to get this picked up by tests, I will move those to a separate PR.

model_servers/common/Makefile.common
```diff
@@ -10,7 +10,7 @@ endif
 
 .PHONY: build
 build:
-	podman build --squash-all --build-arg $(PORT) -t $(IMAGE) . -f base/Containerfile
+	podman build --squash-all --build-arg PORT=$(PORT) -t $(IMAGE) . -f base/Containerfile
```
rhatdan, line 13: This will blow up with PORT is not defined or will at least pass in a PORT="" line.

model_servers/llamacpp_python/base/Containerfile
```diff
@@ -1,6 +1,7 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-52
+ARG PORT
```
rhatdan: ARG PORT=8001

model_servers/llamacpp_python/cuda/Containerfile
```diff
@@ -1,8 +1,10 @@
 FROM quay.io/opendatahub/workbench-images:cuda-ubi9-python-3.9-20231206
+ARG PORT
```
rhatdan: Should this also be PORT=8001?

model_servers/llamacpp_python/vulkan/Containerfile
```diff
@@ -1,5 +1,6 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-52
 USER 0
+ARG PORT
```
rhatdan: PORT=8001?

model_servers/whispercpp/base/Containerfile
```diff
@@ -1,5 +1,5 @@
 FROM registry.access.redhat.com/ubi9/ubi:latest as builder
-
+ARG PORT
```
rhatdan: Ditto

recipes/common/Makefile.common
```diff
@@ -89,6 +89,7 @@ bootc: quadlet
 	  --build-arg APP_IMAGE=$(APP_IMAGE) \
 	  --build-arg SERVER_IMAGE=$(SERVER_IMAGE) \
 	  --build-arg "SSHPUBKEY=$(SSH_PUBKEY)" \
+	  --build-arg PORT=$(PORT) \
```
rhatdan: Is PORT Defined?

I think this would be better as:
```
  $${PORT:+--build-arg PORT=$${PORT}} \
```

recipes/common/Makefile.common
```diff
@@ -101,7 +102,7 @@ bootc: quadlet
 
 .PHONY: bootc-run
 bootc-run:
-	podman run -d --rm --name $(APP)-bootc -p 8080:8501 --privileged \
+	podman run -d --rm --name $(APP)-bootc -p 8080:$(PORT) --privileged \
```
rhatdan, line 105: You need to define a default for PORT somewhere?

recipes/computer_vision/object_detection/model_server/Containerfile
```diff
@@ -1,8 +1,9 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-52
+ARG PORT
 WORKDIR /locallm
 COPY requirements.txt /locallm/requirements.txt
 RUN pip install --upgrade pip && \
     pip install --no-cache-dir --upgrade -r requirements.txt
-COPY object_detection_server.py object_detection_server.py
-EXPOSE 8000
+COPY ../src/object_detection_server.py object_detection_server.py
```
rhatdan: This is illeagle I believe.  You can not refer to a file outside of the context directory.

recipes/computer_vision/object_detection/model_server/Containerfile
```diff
@@ -1,8 +1,9 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-52
+ARG PORT
```
rhatdan: This needs a default,

recipes/natural_language_processing/chatbot/app/Containerfile
```diff
@@ -1,8 +1,9 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-52
+ARG PORT
```
rhatdan: Needs a default

recipes/natural_language_processing/chatbot/bootc/Containerfile
```diff
@@ -6,6 +6,7 @@
 
 FROM quay.io/centos-bootc/centos-bootc:stream9
 ARG SSHPUBKEY
+ARG PORT
```
rhatdan: Needs a default

recipes/natural_language_processing/chatbot/bootc/Containerfile.nocache
```diff
@@ -6,6 +6,7 @@
 
 FROM quay.io/centos-bootc/centos-bootc:stream9
 ARG SSHPUBKEY
+ARG PORT
```
rhatdan: Needs a default

recipes/natural_language_processing/codegen/bootc/Containerfile
```diff
@@ -8,6 +8,7 @@
 
 FROM quay.io/centos-bootc/centos-bootc:stream9
 ARG SSHPUBKEY
+ARG PORT
```
rhatdan: Needs a default

recipes/natural_language_processing/codegen/bootc/Containerfile.nocache
```diff
@@ -6,6 +6,7 @@
 
 FROM quay.io/centos-bootc/centos-bootc:stream9
 ARG SSHPUBKEY
+ARG PORT
```
rhatdan: Needs a default

recipes/natural_language_processing/rag/bootc/Containerfile
```diff
@@ -5,8 +5,8 @@
 # 'make bootc'
 
 FROM quay.io/centos-bootc/centos-bootc:stream9
-
 ARG SSHPUBKEY
+ARG PORT
```
rhatdan: Needs a default

recipes/natural_language_processing/summarizer/bootc/Containerfile
```diff
@@ -6,6 +6,7 @@
 
 FROM quay.io/centos-bootc/centos-bootc:stream9
 ARG SSHPUBKEY
+ARG PORT
```
rhatdan: Needs a default

recipes/natural_language_processing/summarizer/bootc/Containerfile.nocache
```diff
@@ -6,6 +6,7 @@
 
 FROM quay.io/centos-bootc/centos-bootc:stream9
 ARG SSHPUBKEY
+ARG PORT
```
rhatdan: Needs a default

rhatdan:
@MichaelClifford PTAL, are the exposed PORTS above static and defined in Code?

model_servers/common/Makefile.common
Gregory-Pereira: It is defined in every local makefile which calls the common one, so no matter what it will have an argument here. Let me know if you think this is not a good enough solution, but every MS does build successfully with the desired port.

recipes/common/Makefile.common
Gregory-Pereira: Ports are specific to recipe and get defined at the top of those makefiles. This way there will always be a PORT defined.

recipes/computer_vision/object_detection/model_server/Containerfile
Gregory-Pereira: I can change this but all the work for object_detection will live in #236

recipes/common/Makefile.common
Gregory-Pereira: Again I am happy to make this change, but based on the way this is setup, PORT will always be defined, hence conditionally adding the line to the build command doesn't change anything

MichaelClifford:
For the app's the port could change depending on the model server they are connecting to. Our llamacpp_python uses `8001` where as ollama uses `11434`. (We could probably make some changes so they all use the same port though) 

And all the Apps just use `8501` which is streamlit's default. So those don't ever really change.

recipes/computer_vision/object_detection/model_server/Containerfile
```diff
@@ -1,8 +1,9 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-52
+ARG PORT=8001
```
MichaelClifford:
```suggestion
ARG PORT=8000
```

MichaelClifford (changes requested):

MichaelClifford (approved):

recipes/common/Makefile.common
rhatdan, line 105: This should be ${PORT}:80 or something like that.

Gregory-Pereira:
Not sure this is going to end up working. I dusted it off and I am encountering an issue with all the recipe pieces where its very difficult to pass build time ARGs to runtime ARGs in `entrypoint`, which is discussed in this stack overflow thread: https://stackoverflow.com/questions/40902445/using-variable-interpolation-in-string-in-docker. In addition to this, it would require us in the recipe's make file or recipe's common makefile to specify the server port and build it from there, which goes against our intended design. For those reasons I think this is better off dropped and we have to use static ports.

PR #260, opened 2024-04-13
include common directory in tests on path changes

Gregory-Pereira:
Fixing tests, will also use this PR to fix any code changes that correspond to failing tests. 

Broken out from #258.

Changes:

- add a check on jobs in the standard development workflows (all apps, model_servers, model_image_build_push) to not run if there is a `hold-tests` label, I thought this could save us some resources
- fix `codegen`'s `Download model` step due changing make target name
- fix `chromedriver` and `chrome`  downloads failling due to timeout
    - move binaries to `/recipes/common/bin`
- swapped `install` make target in both `recipes/common` and each recipe to be a [double colon rule](https://www.gnu.org/software/make/manual/html_node/Double_002dColon.html#:~:text=4.13%20Double%2DColon%20Rules&text=When%20a%20target%20appears%20in,any%20prerequisites%20of%20that%20rule.) rather than a single colon rule, which allows us to run common and local install commands together

### DIFF
--- a/.github/workflows/chatbot.yaml
+++ b/.github/workflows/chatbot.yaml
@@ -5,12 +5,14 @@ on:
     branches:
       - main
     paths:
+      - ./recipes/common/Makefile.common
       - ./recipes/natural_language_processing/chatbot/**
       - .github/workflows/chatbot.yaml
   push:
     branches:
       - main
     paths:
+      - ./recipes/common/Makefile.common
       - ./recipes/natural_language_processing/chatbot/**
       - .github/workflows/chatbot.yaml
 
@@ -22,6 +24,7 @@ env:
 
 jobs:
   build-and-push-image:
+    if: "!contains(github.event.pull_request.labels.*.name, 'hold-tests')"
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/recipes/common/Makefile.common
+++ b/recipes/common/Makefile.common
@@ -31,7 +31,7 @@ endif
 CHROMEDRIVER_VERSION := 103.0.5060.53
 CHROMEDRIVER_MIRROR := https://chromedriver.storage.googleapis.com
 CHROMEDRIVER_DOWNLOAD_PATH := 
-CHROMEDRIVER_INSTALLATION_PATH ?=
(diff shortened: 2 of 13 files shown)


Gregory-Pereira:
Tests pass, going to add the `hold-tests` label and push an empty commit to see if it works.

Gregory-Pereira:
Can confirm that the labels to skip tests work. Removing `hold-tests` label

.github/workflows/chatbot.yaml
```diff
@@ -22,6 +24,7 @@ env:
 
 jobs:
   build-and-push-image:
+    if: "!contains(github.event.pull_request.labels.*.name, 'hold-tests')"
```
lmilbaum, line 27: Not sure folks would remember to use this trick

recipes/common/Makefile.common
```diff
@@ -129,9 +129,9 @@ bootc-image-builder: bootc
 install-chromedriver:
 	@if [[ -z "$(LOCAL_CHROMEDRIVER_EXISTS)" ]]; then \
 		if [[ -n "$(UNZIP_EXISTS)" ]]; then \
-			curl -sLO $(CHROMEDRIVER_MIRROR)/$(CHROMEDRIVER_VERSION)/$(CHROMEDRIVER_DOWNLOAD_PATH); \
+			curl --max-time 300 --connect-timeout 60 -sLO $(CHROMEDRIVER_MIRROR)/$(CHROMEDRIVER_VERSION)/$(CHROMEDRIVER_DOWNLOAD_PATH); \
```
lmilbaum, line 132: Nice. I was looking into a retry option. That is a better option to solve those flaky results.

lmilbaum (approved):
LGTM

.github/workflows/chatbot.yaml
Gregory-Pereira: I was going to have a discussion with the team about it.  I know I would use it because sometime I just want to see the diff for review purposes and dont want to waist the resources. If your worried about visibility maybe we could tag every PR with `hold-tests` to begin with and then use a `/ok-to-test` similar to prow?

Issue #237, opened 2024-04-11
confusing recipe descriptions for Podman Desktop users


vrothberg:
We received user feedback that the recipe descriptions can be confusing.  Let's take the ChatBot as an example (see screenshot) below.  The description does not target the desktop use case but the terminal/CLI one.  So there is some friction.

Since the AI Lab Recipes target both use cases, I _guess_ that the READMEs should first target the desktop use case and then the CLI use with clear descriptions of the target audience?

![Screenshot 2024-04-11 at 10 59 31](https://github.com/containers/podman-desktop-extension-ai-lab/assets/11679875/f8f50e30-c008-4b4c-a595-09c21d4a0485)

vrothberg:
I think that the latest README goes into these details: https://github.com/containers/ai-lab-recipes/tree/main/recipes/natural_language_processing/chatbot

vrothberg:
At least for the ChatBot.  I guess we need to add information about Podman Desktop and AI Lab to all READMEs.

@sallyom, @rhatdan, what do you think?

rhatdan:
SGTM, It would be good to get more eyes on the README.  I was playing around with documenting the Makefile targets and variables but I think there is too much cut and paste code going on.  But markdown is not great about sharing code.

rhatdan:
I think we should go through the chatbot README.md, get it up to speed, then work on getting it distributed to the other recipes.

vrothberg:
> I think we should go through the chatbot README.md, get it up to speed, then work on getting it distributed to the other recipes.

Great idea!

rhatdan:
If you can get images from podman desktop to match the podman command line, we can scatter them throughout the README.md file.

vrothberg:
> If you can get images from podman desktop to match the podman command line, we can scatter them throughout the README.md file.

Coming from the meeting:  I guess "images" means screenshots, right?

rhatdan:
Yes

vrothberg:
Approaching it from a new angle, I don't think we should add screenshots.  The content of this README is displayed directly inside of Podman Desktop, so screenshots may confuse the user and they easily get outdated.

vrothberg:
Closing as most NLP recipes are now in sync.  We need to keep an eye out when updating other READMEs to make sure they use the same description as the ChatBot one.

PR #267, opened 2024-04-15
chatbot README: a couple of refinements

vrothberg:
Part-of: #237

@rhatdan @sallyom PTAL


vrothberg:
If this gets in, I will open a follow-up PR to update the other recipe READMEs as well.

rhatdan:
LGTM

PR #270, opened 2024-04-15
Fix ARCH= option for bootc images.

rhatdan:
If user overrides the arch of the image, then the podman pull lines in the bootc container need to follow the TARGETARCH and pull the correct image.


rhatdan:
@mrguitar @sallyom PTAL

Gregory-Pereira (approved):
/lgtm

PR #256, opened 2024-04-12
fix: update object_detection recipe for be compatible with AI Lab

jeffmaury:
Fixes #220

PR #273, opened 2024-04-15
move object_detection MS to MS dir

Gregory-Pereira:
/cc @MichaelClifford


model_servers/object_detection_python/src/object_detection_server.py
```diff
@@ -9,7 +9,9 @@
 
 
 app = FastAPI()
-model = os.getenv("MODEL_PATH", default="facebook/detr-resnet-101")
+models_path = os.getenv("MODELS_PATH", default="/app/models")
+model_name = os.getenv("MODEL_NAME", default="detr-resnet-101")
+model = os.path.abspath(f"{models_path}/{model_name}")
```
MichaelClifford: I don't think that this is compatible with the way AI Lab will expect the model paths to be. Can you revert to the original. Jeff is changing this file in #256 as well.

MichaelClifford (changes requested):

model_servers/object_detection_python/src/object_detection_server.py
```diff
@@ -9,7 +9,9 @@
 
 
 app = FastAPI()
-model = os.getenv("MODEL_PATH", default="facebook/detr-resnet-101")
+models_path = os.getenv("MODELS_PATH", default="/models")
+model_name = os.getenv("MODEL_NAME", default="facebook/detr-resnet-101")
+model = os.path.abspath(f"{models_path}/{model_name}")
```
MichaelClifford: Please revert this file.  This `model` path will not  be compatible with changes being made in #256

MichaelClifford (approved):
LGTM

PR #275, opened 2024-04-16
Add model download utility

MichaelClifford:
Added `download_hf_models.py` to have a single utility for downloading models from huggingface.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ port_check.lock
 models/*
 !models/Makefile
 !models/README.md
+!models/download_hf_models.py
 convert_models/converted_models
 recipes/common/bin/*
 */.venv/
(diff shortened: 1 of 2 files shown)


rhatdan:
Is there any chance of a competitor to Hugging Face emerging?  Is it likely the same interface work with a new one?

MichaelClifford:
@rhatdan yes, there is a chance a competitor to huggingface emerges, but I'm not sure how likely it would be in the next year.  This code uses the huggingface's python package, so I don't think it would be compatible with other model sources. Should a competitor arise, we'd need to address that.

That said, the goal here is an attempt at setting some standard on how models get stored throughout all the recipes since there can be differences between GGUF and other model types. For now, since we are only using huggingface, this PR should enforce a `models/<MODEL_NAME>/*` directory structure on all models regardless of their type.

Should another provider need to be supported, we'd just want ensure we are storing model on our end in a uniform way.

rhatdan:
LGTM

PR #293, opened 2024-04-21
Fix document chunking in summarizer

MichaelClifford:
Noticed an issue in the summarizer app where text files larger than the models context length where not being handle correctly.

This PR updates `chunk_text()` to make sure all text chunks are well within the token limit, and that the list of of text chunks is updated correctly. 

Also removed `estimate_tokens()` as it was no longer needed with these new changes. 

Also, set the model temperature value to `0` as that should lead to less creative outputs, and better results for summarization tasks.


rhatdan:
LGTM

Issue #69, opened 2024-02-29
Text to Json Recipe


MichaelClifford:
A recipe that does the following:

Given a small set of json schemas (Ok to start with 1), create an LLM application where a user inputs their responses to the fields of the json schema using unstructured natural language and the model returns a correctly populated json.

A simple example to illustrate how this should work would be placing an order at a fast food restaurant. The user simply states their order, and the LLM generates the appropriate json that can be submitted to their oder management system.

PR #294, opened 2024-04-21
WIP, pass one at populate-json recipe

Gregory-Pereira:
Progress towards #69


rhatdan:
@MichaelClifford PTAL

Gregory-Pereira:
This is just a general implementation so far. I found for performance reasons we could not use mistral-7b, as such I will be picking up some of Micahels work to convert TF and pytorch models to GGUF and vice versa

PR #299, opened 2024-04-22
rfe(bootc-image-builder): Support "--chown" option

YvanY0:
bootc-image-builder requests root permission, which means the generated disk image is own by the root user. Helpfully, bib provides the "--chown" option to help us change the owner of the output directory. This make user easy to custom UID:GID of the disk and use it later.

### DIFF
--- a/recipes/common/Makefile.common
+++ b/recipes/common/Makefile.common
@@ -10,6 +10,8 @@ SSH_PUBKEY ?= $(shell cat ${HOME}/.ssh/id_rsa.pub;)
 BOOTC_IMAGE ?= quay.io/$(REGISTRY_ORG)/${APP}-bootc:latest
 BOOTC_IMAGE_BUILDER ?= quay.io/centos-bootc/bootc-image-builder
 DISK_TYPE ?= qcow2
+DISK_USER ?= $(shell id -u)
+DISK_GROUP ?= $(shell id -g)
 FROM ?=
 ARCH ?=
 CONTAINERFILE ?= Containerfile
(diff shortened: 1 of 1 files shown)


YvanY0:
With this, I can use `sudo make bootc-image-builder DISK_USER=$(id -u) DISK_GROUP=$(id -g)` to generate the disk image and have it owned by the current user. So user can directly use it to launch the VM.

Ref: https://github.com/osbuild/bootc-image-builder#detailed-description-of-optional-flags

recipes/common/Makefile.common
```diff
@@ -15,6 +15,10 @@ ARCH ?=
 CONTAINERFILE ?= Containerfile
 GRAPH_ROOT=$(shell podman info --format '{{ .Store.GraphRoot }}')
 
+DISK_USER ?=
```
rhatdan: Why not default these to
 DISK_USER=$(id -u) DISK_GROUP=$(id -g)

YvanY0: Hello @rhatdan, with sudo, the makefile is executed by the root user, so the default value in the Makefile is meaningless. We want to change the disk image to non-root owner, if set default value, bib will add `--chown 0:0`.

Here is an example:
```bash
$ cat Makefile
DISK_USER ?= $(shell id -u)
DISK_GROUP ?= $(shell id -g)

.PHONY: show-ids

show-ids:
	@echo "UID: $(DISK_USER)"
	@echo "GID: $(DISK_GROUP)"

$ make
UID: 1001
GID: 1001

$ sudo make
UID: 0
GID: 0

$ sudo make DISK_USER=$(id -u) DISK_GROUP=$(id -g)
UID: 1001
GID: 1001
```

rhatdan: Right but for non root users, they don't have to think about it and the right thing happens.
For rootful users this becomes a noop chown root to root.

YvanY0: OK, so you mean even podman chown the disk by `0:0`, We also explicitly specify it? I can accept this, let me change my code, and see if other people have concerns.

The current design is only when `DISK_USER` + `DISK_GROUP` are present, then `--chown` will be added.

rhatdan:
LGTM

PR #302, opened 2024-04-22
add PDF reading to summarizer

MichaelClifford:
* Updated Summarizer to read pdf files
* Placed limit on "upload widget" so only pdf or txt files can be uploaded.
* Added PDF version of `fake_meeting.txt` for testing. 

_Note: PDF's can now be read. However, while testing I found that more complex PDF's such as research papers can cause some issues (the summary will crash). I believe that this has something to do with the high occurrence of non-standard text (math equations, tables, graphs, etc) and the way we are estimating tokens for chunking. There should be a follow up issue for addressing the handling more complex document types._


recipes/natural_language_processing/summarizer/app/summarizer.py
```diff
@@ -44,14 +45,30 @@ def chunk_text(text):
         text_list = text_list[chunk_size:]
     return chunks
 
+def read_file(file):
+    file_type = file.type
+    
+    if file_type == "application/pdf":
+        with open("tmp_pdf", "wb") as f:
```
Gregory-Pereira: Shouldnt this use the file_name:
```suggestion
        with open(file_name, "wb") as f:
```
Also I am curious why we need to `f.write` before we can call it in with the loader

MichaelClifford: I don't love that implementation either, happy to see if we can get around it. But for now the PDF loader we're using seems to want to read in a file from disk. So we write the streamlit file object to disk temporarily so we can then read it using the PDF loader.

Gregory-Pereira: I see, that makes sense. We can iterate as we go but that is more than acceptable for a first pass, I was just wondering what the effect of this step was.

recipes/natural_language_processing/summarizer/app/summarizer.py
```diff
@@ -44,14 +45,30 @@ def chunk_text(text):
...
+        os.remove(file.name) 
+    
+    if file_type == "text/plain":
+        text = file.read().decode()   
+    
+    return text
+
 st.title("🔎 Summarizer")
-file = st.file_uploader("Upload file")
+file = st.file_uploader("Upload file",type=[".txt",".pdf"])
```
sallyom, line 65: if I upload a text file without the `.txt` extension, it will fail, right? If so, it might be better to document "Upload text file or pdf" and not enforce there, wdyt?

MichaelClifford: I don't love that idea. Mainly because we use `file.type` to determine how we handle opening the file and that value gets determined by the extension. We could, but I think its probably simpler for users to see that they need to add ".txt" to their file names, then it is to manage the open-ended errors that could occur by allowing any file type to be uploaded in the ui.  

What do others think?

rhatdan:
I am going to merge, we can continue to discuss and get user feedback

PR #303, opened 2024-04-22
Fixup download-models

rhatdan:
The models/README.md described a bunch of download-model flags
that did not exist or were defined in different makefiles. This
PR removes the non-existing targets and adds the defined targets
into the models/Makefile and referenced from the others.

### DIFF
--- a/model_servers/whispercpp/Makefile
+++ b/model_servers/whispercpp/Makefile
@@ -17,4 +17,4 @@ all: build download-model-whisper-small run
 .PHONY: download-model-whisper-small # small .bin model type testing
 download-model-whisper-small:
 	cd ../../models && \
-	make MODEL_NAME=ggml-small.bin MODEL_URL=https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-small.bin -f Makefile download-model
+	make download-model-whisper-small
--- a/models/Makefile
+++ b/models/Makefile
@@ -24,3 +24,11 @@ download-model-granite:
 .PHONY: download-model-merlinite
 download-model-merlinite:
 	$(MAKE) MODEL_URL=https://huggingface.co/instructlab/merlinite-7b-lab-GGUF/resolve/main/merlinite-7b-lab-Q4_K_M.gguf MODEL_NAME=merlinite-7b-lab-Q4_K_M.gguf download-model
+
+.PHONY: download-model-whisper-small # small .bin model type testing
+download-model-whisper-small:
+	$(MAKE) MODEL_NAME=ggml-small.bin MODEL_URL=https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-small.bin download-model
+
+.PHONY: download-model-mistral
+download-model-mistral:
+	$(MAKE) MODEL_NAME=mistral-7b-instruct-v0.1.Q4_K_M.gguf MODEL_URL=https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf download-model
--- a/recipes/common/Makefile.common
+++ b/recipes/common/Makefile.common
@@ -64,11 +64,6 @@ MISTRAL_MODEL_URL := https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GG
 
 MODEL_NAME ?= $(MISTRAL_MODEL_NAME)
 
-.PHONY: download-model-mistral
-download-model-mistral:
-	cd ../../../models && \
-	make MODEL_NAME=mistral-7b-instruct-v0.1.Q4_K_M.gguf MODEL_URL=https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf -f  Makefile download-model
-
 .PHONY: install
 install::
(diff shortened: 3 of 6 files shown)


rhatdan:
@sallyom @MichaelClifford @Gregory-Pereira PTAL

models/Makefile
```diff
@@ -19,8 +19,16 @@ download-model:
 
 .PHONY: download-model-granite
 download-model-granite:
-	$(MAKE) MODEL_URL=https://huggingface.co/instructlab/granite-7b-lab-GGUF/resolve/main/granite-7b-lab-Q4_K_M.gguf MODEL_NAME=granite-7b-lab-Q4_K_M.gguf download-model
+	$(MAKE) download-model MODEL_URL=https://huggingface.co/instructlab/granite-7b-lab-GGUF/resolve/main/granite-7b-lab-Q4_K_M.gguf MODEL_NAME=granite-7b-lab-Q4_K_M.gguf download-model
```
Gregory-Pereira: duplicated `download-model` command

models/Makefile
```diff
@@ -19,8 +19,16 @@ download-model:
...
 
 .PHONY: download-model-merlinite
 download-model-merlinite:
-	$(MAKE) MODEL_URL=https://huggingface.co/instructlab/merlinite-7b-lab-GGUF/resolve/main/merlinite-7b-lab-Q4_K_M.gguf MODEL_NAME=merlinite-7b-lab-Q4_K_M.gguf download-model
+	$(MAKE) download-model MODEL_URL=https://huggingface.co/instructlab/merlinite-7b-lab-GGUF/resolve/main/merlinite-7b-lab-Q4_K_M.gguf MODEL_NAME=merlinite-7b-lab-Q4_K_M.gguf download-model
```
Gregory-Pereira: Same here, duplicate `download-model` command

recipes/common/Makefile.common
```diff
@@ -65,7 +65,7 @@ MODEL_NAME ?= $(MISTRAL_MODEL_NAME)
 .PHONY: download-model-mistral
 download-model-mistral:
 	cd ../../../models && \
-	make MODEL_NAME=mistral-7b-instruct-v0.1.Q4_K_M.gguf MODEL_URL=https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf -f  Makefile download-model
+	make download-model-mistral
```
Gregory-Pereira: Where is this being referenced, common does not pull in any other makefiles so shouldn't this be an undefined target?

Gregory-Pereira (commented):
Some questions:

models/Makefile
rhatdan: Oops

recipes/common/Makefile.common
rhatdan: Ok removed the target.

recipes/common/Makefile.common
```diff
@@ -62,11 +62,6 @@ MISTRAL_MODEL_URL := https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GG
 
 MODEL_NAME ?= $(MISTRAL_MODEL_NAME)
 
-.PHONY: download-model-mistral
-download-model-mistral:
-	cd ../../../models && \
-	make MODEL_NAME=mistral-7b-instruct-v0.1.Q4_K_M.gguf MODEL_URL=https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf -f  Makefile download-model
```
Gregory-Pereira, line 70: I think this should be:
```Makefile
.PHONY: download-model-mistral
download-model-mistral:
	$(MAKE) -f  ../../../models/Makefile download-model-mistral
```
We want both `/model_servers/common` and `/recipes/common` to use the `/models` downloads IIUC

model_servers/whispercpp/Makefile
```diff
@@ -17,4 +17,5 @@ all: build download-model-whisper-small run
 .PHONY: download-model-whisper-small # small .bin model type testing
 download-model-whisper-small:
 	cd ../../models && \
-	make MODEL_NAME=ggml-small.bin MODEL_URL=https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-small.bin -f Makefile download-model
+	make download-model-whisper-small
```
Gregory-Pereira, line 20: I think we remove this entire make target and instead in the `model_servers/common/Makefile.common` we add:
```Makefile
.PHONY: download-model-mistral
download-model-mistral:
	$(MAKE) -f  ../../models/Makefile download-model-mistral
```
This allows us to leverage common to provide all model options to all servers.

Gregory-Pereira (commented):
A little more restructuring:

recipes/common/Makefile.common
rhatdan: I guess the question is where do we want the models downloaded to, if we do this I think the models will be downloaded in the model_servers or recipes directory?  Is that the correct place to download them.

I was surprised to find all of these huge files scattered in subdirs. Where I would be less surprised to find them in the models subdir.

rhatdan: cd ../../../modles; $(MAKE) download-model-mistral

Would always download in the models dir.

MichaelClifford: The purpose of the models dir, is to house all the models. There should not be models anywhere else in the repo.

MichaelClifford (approved):
LGTM

PR #300, opened 2024-04-22
adding merlinite and granite models

Gregory-Pereira:
Addresses: https://redhat-internal.slack.com/archives/C06S75ZF9JT/p1713792833949739
/cc @rhatdan @sallyom

PR #304, opened 2024-04-22
add model workflows

sallyom:
adding merlinite & granite model workflow

### DIFF
--- a/.github/workflows/model_image_build_push.yaml
+++ b/.github/workflows/model_image_build_push.yaml
@@ -9,9 +9,15 @@ on:
 env:
   REGISTRY: quay.io/ai-lab
   MODEL_CONVERTER_IMAGE_NAME: model-converter
+  MERLINITE_MODEL_IMAGE_NAME: merlinite-7b-lab
+  MERLINITE_LABEL: Q4_K_M
+  MERLINITE_MODEL_URL: https://huggingface.co/instructlab/merlinite-7b-lab-GGUF/resolve/main/merlinite-7b-lab-Q4_K_M.gguf
   MISTRAL_MODEL_IMAGE_NAME: mistral-7b-instruct
   MISTRAL_LABEL: v0.1.Q4_K_M.gguf
   MISTRAL_MODEL_URL: https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf
+  GRANITE_MODEL_IMAGE_NAME: granite-7b-lab
+  GRANITE_LABEL: Q4_K_M
+  GRANITE_MODEL_URL: https://huggingface.co/instructlab/granite-7b-lab-GGUF/resolve/main/granite-7b-lab-Q4_K_M.gguf
 
 jobs:
   build-and-push-model-converter-image:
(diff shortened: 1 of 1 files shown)


.github/workflows/model_image_build_push.yaml
```diff
@@ -9,9 +9,15 @@ on:
...
   MODEL_CONVERTER_IMAGE_NAME: model-converter
+  MERLINITE_MODEL_IMAGE_NAME: merlinite-7b-lab
+  MERLINITE_LABEL: Q4_K_M
+  MERLINITE_MODEL_URL: https://huggingface.co/instructlab/merlinite-7b-lab-GGUF/resolve/main/merlinite-7b-lab-Q4_K_M.gguf
   MISTRAL_MODEL_IMAGE_NAME: mistral-7b-instruct
   MISTRAL_LABEL: v0.1.Q4_K_M.gguf
   MISTRAL_MODEL_URL: https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf
+  GRANITE_MODEL_IMAGE_NAME: granite-7b-lab
+  GRANITE_LABEL: Q4_K_M
+  GRANITE_MODEL_URL: https://huggingface.co/instructlab/granite-7b-lab-GGUF/resolve/main/granite-7b-lab-Q4_K_M.gguf
```
Gregory-Pereira, line 20: We should do these using our make targets for greater test coverage via a matrix rather than hardcoding values into the ENV, see [example in ms workflows](https://github.com/containers/ai-lab-recipes/blob/main/.github/workflows/model_servers.yaml#L27-L46)
```suggestion
```

.github/workflows/model_image_build_push.yaml
```diff
@@ -61,6 +67,108 @@ jobs:
...
+      - name: Install qemu dependency
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y qemu-user-static
+
+      - name: Build granite model image
+        id: build_granite_model_image
+        uses: redhat-actions/buildah-build@v2
+        with:
+          image: ${{ env.GRANITE_MODEL_IMAGE_NAME }}
```
Gregory-Pereira, line 96: Based on my above suggestion this will move to a matrix based approach, ex [ms model downloads via matrix](https://github.com/containers/ai-lab-recipes/blob/main/.github/workflows/model_servers.yaml#L79-L81)

.github/workflows/model_image_build_push.yaml
```diff
@@ -61,6 +67,108 @@ jobs:
...
+          tags: latest
+          platforms: linux/amd64, linux/arm64
+          context: models
+          labels: |
+            ${{ env.GRANITE_LABEL }}
```
Gregory-Pereira, line 101: Another env --> matrix refactor

.github/workflows/model_image_build_push.yaml
```diff
@@ -61,6 +67,108 @@ jobs:
...
+          build-args: |
+            MODEL_URL=${{ env.GRANITE_MODEL_URL }}
```
Gregory-Pereira, line 103: Another env --> matrix refactor

.github/workflows/model_image_build_push.yaml
```diff
@@ -61,6 +67,108 @@ jobs:
...
+        with:
+          image: ${{ env.MERLINITE_MODEL_IMAGE_NAME }}
+          tags: latest
+          platforms: linux/amd64, linux/arm64
+          context: models
+          labels: |
+            ${{ env.MERLINITE_LABEL }}
+          build-args: |
+            MODEL_URL=${{ env.MERLINITE_MODEL_URL }}
+          containerfiles: ./models/Containerfile
```
Gregory-Pereira, line 155: Same as discussed above

Gregory-Pereira (commented):
@sallyom firstly this is awesome stuff. Also my appologies this should have been included with #300 so thanks for adding this! However I would like to move to a matrix based approach with all our workflows as it is easier and safer to extend. You dont have to take the following suggestions but I highly recommend:

Gregory-Pereira:
One thing you should also do is include is more targeted `on` policy to specify what we want to test for. Here is an of what this looks like for our [chatbot workflow](https://github.com/containers/ai-lab-recipes/blob/main/.github/workflows/chatbot.yaml#L4-L17)

It should look something like this:
```yaml
on:
  schedule: # schedule the job to run at 12 AM daily
   - cron: '0 0 * * *'
   
  pull_request:
    branches:
      - main
    paths:
      - ./models/Makefile
      - .github/workflows/model_image_build_push.yaml
  push:
    branches:
      - main
    paths:
      - ./models/Makefile
      - .github/workflows/model_image_build_push.yaml
```

.github/workflows/model_image_build_push.yaml
sallyom: matrix approach works for me! it'll be much better

Gregory-Pereira:
It was determined we want to push ahead and get these images in Quay ASAP. I will follow on with a refactor PR adressing the concerns laid out above.

Gregory-Pereira (approved):
/lgtm
@Gregory-Pereira to refactor after images get created

sallyom:
@Gregory-Pereira is going to refactor this and make it 1000% better, I'm merging this as/is to populate these images

PR #307, opened 2024-04-22
fix convert_models build with build path permissions (temporary)

Gregory-Pereira:
/cc @sallyom @MichaelClifford this was currently breaking the push model image workflow

### DIFF
--- a/.github/workflows/model_image_build_push.yaml
+++ b/.github/workflows/model_image_build_push.yaml
@@ -4,6 +4,19 @@ on:
   schedule: # schedule the job to run at 12 AM daily
    - cron: '0 0 * * *'
   
+  pull_request:
+    branches:
+      - main
+    paths:
+      - ./convert_models/**
+      - .github/workflows/model_image_build_push.yaml
+  push:
+    branches:
+      - main
+    paths:
+      - ./convert_models/**
+      - .github/workflows/model_image_build_push.yaml
+
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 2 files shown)


rhatdan:
Where did /opt/app/root come from?

Gregory-Pereira:
That is the default (app-root) of the parent container: `FROM registry.access.redhat.com/ubi9/python-311:1-52.1712567218`. In the workflows were solving for a permissions write error: 
```Container build log
STEP 2/7: WORKDIR /converter
STEP 3/7: RUN git clone https://github.com/ggerganov/llama.cpp.git
fatal: could not create work tree dir 'llama.cpp': Permission denied
```

Gregory-Pereira:
This results from the following:
```bash
 $ podman run --rm -it registry.access.redhat.com/ubi9/python-311:1-52.1712567218 /bin/bash
(app-root) cd /
(app-root) ls
afs  bin  boot	dev  etc  help.1  home	lib  lib64  lost+found	media  mnt  opt  proc  root  run  sbin	srv  sys  tmp  usr  var
(app-root) git clone https://github.com/ggerganov/llama.cpp.git
fatal: could not create work tree dir 'llama.cpp': Permission denied
(app-root) mkdir /converter
mkdir: cannot create directory ‘/converter’: Permission denied
```

sallyom (approved):
LGTM, thank you!

Gregory-Pereira:
Since you approve sally I think we can merge this, as we are hotfixing the integration tests anyway, more PRs can follow to cleanup.

sallyom:
Try with the original Containerfile, but remove the slash `/` from `/converter`

PR #308, opened 2024-04-22
try non protected build with path changes

Gregory-Pereira:
@sallyom's suggested sollution


sallyom:
something doesn't sit right with this-I'm going to go back to why it was originally failing... also, can you give the absolute path to converter? You exec-d in and found the current workdir of the base image, right?
`WORKDIR /opt/app-root/src/converter`

Gregory-Pereira:
Confirmed this works, linting and merging.

PR #309, opened 2024-04-22
temporarily remove model-converter image build

sallyom:
There's an issue with the model_converter image job that is blocking the model image builds. @Gregory-Pereira is going to add this image back in with his refactor PR 🙇 #308


sallyom:
self-merging, unblocks adding model images to quay.io/ai-lab

PR #315, opened 2024-04-23
fix codegen quadlet variables

sallyom:
fixes some issues with codegen recipe


sallyom:
closing, adding these to the PR that adds the code model: https://github.com/containers/ai-lab-recipes/pull/314

PR #316, opened 2024-04-23
Accelerated container image for AMD

kwozyman:
I've added the first bootc bare metal accelerated bootc container image and it's README.

### DIFF
--- a/training/Makefile
+++ b/training/Makefile
@@ -1,0 +1,13 @@
+FROM ?= "quay.io/centos-bootc/centos-bootc:stream9"
+
+REGISTRY ?= quay.io
+REGISTRY_ORG ?= ai-lab
+IMAGE_NAME ?= amd
+IMAGE_TAG ?= latest
+
+CONTAINER_TOOL ?= podman
+CONTAINER_TOOL_EXTRA_ARGS ?=
+
+.PHONY: amd
+amd:
+	make -C amd/ image
--- a/training/amd/Containerfile
+++ b/training/amd/Containerfile
@@ -1,0 +1,6 @@
+FROM quay.io/centos-bootc/centos-bootc:stream9
+
+ADD rocm.repo /etc/yum.repos.d/rocm.repo
+
+RUN curl -L https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm -o /tmp/epel-release-latest-9.noarch.rpm && rpm -ivh /tmp/epel-release-latest-9.noarch.rpm && rm -f /tmp/epel-release-latest-9.noarch.rpm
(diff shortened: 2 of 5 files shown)


training/Makefile
```diff
@@ -0,0 +1,13 @@
...
+REGISTRY_ORG ?= ai-lab
+IMAGE_NAME ?= amd
+IMAGE_TAG ?= latest
+
+CONTAINER_TOOL ?= podman
+CONTAINER_TOOL_EXTRA_ARGS ?=
+
+.PHONY: amd
+amd:
+	make -C amd/ image
```
rhatdan, line 13: I think you would need to change this to
make -C amd/ image FROM=$(FROM) REGISTRY=$(REGISTRY) REGISTRY_ORG=$(REGISTRY_ORG)...

training/amd/Makefile
```diff
@@ -0,0 +1,18 @@
...
+
+CONTAINER_TOOL ?= podman
+CONTAINER_TOOL_EXTRA_ARGS ?=
+
+.PHONY: image
+image:
+	"${CONTAINER_TOOL}" build \
+		--file Containerfile \
+		--tag "${REGISTRY}/${REGISTRY_ORG}/${IMAGE_NAME}:${IMAGE_TAG}" \
+		--build-arg BASEIMAGE=${FROM} \
```
rhatdan: In the other repos we have been using 

--from $(FROM) rather then --build-arg.

training/amd/Containerfile
```diff
@@ -0,0 +1,8 @@
+ARG BASEIMAGE="quay.io/centos-bootc/centos-bootc:stream9"
+
+FROM ${BASEIMAGE}
```
rhatdan: I would prefer to use `FROM quay.io/centos-bootc/centos-bootc:stream9` and use `--from $(FROM) in the Makefile like we do in the rest of the repo.

rhatdan:
Looks good, just a couple of nits.

training/Makefile
kwozyman: I think these get inherited as environment variables. At least on my Fedora 39 a `make amd REGISTRY=foo` gets me the correct name just with the above

training/amd/Containerfile
kwozyman: Done

training/amd/Makefile
kwozyman: Done

training/Makefile
rhatdan: If they do that is great, ignore my comment.

rhatdan:
@kwozyman you need to sign your commits, and then we can merge.

rhatdan:
LGTM

PR #317, opened 2024-04-23
fix typo in model image names on push

sallyom:
images are being built with the correct name, but pushed as `quay.io/ai-lab/quay.io/ai-lab/image-name` - this fixes it


sallyom:
merging, last fix to model-image builds 🤞

PR #318, opened 2024-04-23
after 10 failed attempts, finally updating the model-image workflow the right way

sallyom:
@Gregory-Pereira


rhatdan:
LGTM
@lmilbaum PTAL

PR #319, opened 2024-04-23
Update model download example in models README.md

MichaelClifford:
Download example was referencing a model that does not have an Apache or MIT licenses. Changed for one that does.


sallyom (approved):
LGTM

rhatdan:
LGTM

PR #320, opened 2024-04-23
remove extra registry in image push

sallyom:
hoping this is it 🙏 @lmilbaum @Gregory-Pereira ptal


Gregory-Pereira (approved):
Sally is victorious! 
/lgtm

Issue #280, opened 2024-04-17
feat(audio_to_text): convert audio to the expected format


axel7083:
As raised by @lstocchi in https://github.com/containers/podman-desktop-extension-ai-lab/pull/869#pullrequestreview-2005320708 user may try to use some format which are not compatible.

As `ffmpeg` is available, the `whisper_client.py` could check for the format of the source audio and convert it to the expected format before sending it.

cc @MichaelClifford

MichaelClifford:
Similar to https://github.com/containers/ai-lab-recipes/issues/64.  wIll close 64 in favor of this one.

PR #326, opened 2024-04-23
Update whisper model recipe to convert audio files

hemajv:
Update the audio to text recipe to check for the format of the source audio and convert it to the expected format

(Addresses #280)


rhatdan:
LGTM
@MichaelClifford PTAL

MichaelClifford (approved):

PR #330, opened 2024-04-24
set summarizer write to tempfile

MichaelClifford:
Follow up to #302 

Noticed summarizer was having write issues in container. Using the `/tmp/` directory should address it.

### DIFF
--- a/recipes/natural_language_processing/summarizer/app/summarizer.py
+++ b/recipes/natural_language_processing/summarizer/app/summarizer.py
@@ -4,6 +4,7 @@ from langchain.prompts import PromptTemplate
 from langchain_community.callbacks import StreamlitCallbackHandler
 from langchain_community.document_loaders import PyPDFLoader
 import streamlit as st
+import tempfile
 import requests
 import time
 import os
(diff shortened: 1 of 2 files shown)


Gregory-Pereira:
Similar issue encountered in: #307 but this solves it a better way

Gregory-Pereira (approved):
/lgtm

rhatdan:
There is security risks in writing to tmp, can you use a tmpfile name rather then a hard coded file name?

MichaelClifford:
Thanks @rhatdan Changed it to use a tempfile instead.

recipes/natural_language_processing/summarizer/app/summarizer.py
```diff
@@ -49,12 +50,12 @@ def read_file(file):
     file_type = file.type
     
     if file_type == "application/pdf":
-        with open(file.name, "wb") as f:
+        temp = tempfile.NamedTemporaryFile()
```
rhatdan, line 53: You still need to remove this when you are done with it?

MichaelClifford: When I tested it the tempfile deleted after the file was no longer being used.

rhatdan: Is this some magic in Python?

rhatdan:
```
Use the [tempfile module](http://docs.python.org/2/library/tempfile.html); it creates temporary files that auto-delete.

From the [tempfile.NamedTemporaryFile() documentation](http://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile):

    If delete is true (the default), the file is deleted as soon as it is closed.
```
Cool

rhatdan:
LGTM

PR #332, opened 2024-04-24
Better separation between CentOS and RHEL

kwozyman:
I've made a selector based on /etc/os-release for enabling CRB repo for CentOS. For RHEL, the CRB repo should be enabled on the build host.


rhatdan:
LGTM

PR #334, opened 2024-04-24
Add Milvus database compatibility with the RAG recipe

Shreyanand:
This PR adds the following changes:

- Adds a shell script to deploy a standalone Milvus container on the local system
- Seperate the vector db management logic from rag_chat.py file
- Add a logic to switch between Milvus or Chromadb depending on what is being used

@MichaelClifford I need to add documentation around this but I have tested it on my system and it works. What should be the next things to check to make sure this is polished?


rhatdan:
@Shreyanand you need to sign your commits.

git commit -a --amend -s
git push --force

rhatdan:
@MichaelClifford PTAL

recipes/natural_language_processing/rag/app/manage_vectordb.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+    # def has_collection(self):
+    #     # Implement logic to check if the VectorDB has a collection
+    #     print(f"Checking if collection {self.collection_name} exists in VectorDB...")
+    #     if self.vector_vendor == "chromadb":
+    #         hc = collection.count() < 1
+    #     elif self.vector_vendor == "milvus":
+    #         hc =
+    #     return hc
```
MichaelClifford: please remove this dead code

recipes/natural_language_processing/rag/app/manage_vectordb.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+    def create_retriever(self):
+        # Implement logic to create and return a retriever object from the VectorDB
+        print("Creating retriever object from VectorDB...")
```
MichaelClifford: Remove this, doesn't appear to do anything.

recipes/natural_language_processing/rag/app/manage_vectordb.py
```diff
@@ -0,0 +1,98 @@
...
+                db = Milvus(
+                    e,
+                    collection_name=self.collection_name,
+                    connection_args={"host": self.host, "port": self.port},
+                )
+        return db
+    
+    def clear_db(self):
+        # Implement logic to clear the entire VectorDB
+        print(f"Clearing the entire VectorDB...")
```
MichaelClifford:
```suggestion
        print(f"Clearing VectorDB...")
```

recipes/natural_language_processing/rag/app/manage_vectordb.py
```diff
@@ -0,0 +1,98 @@
...
+                print("DB already populated")
+                db = Milvus(
+                    e,
+                    collection_name=self.collection_name,
+                    connection_args={"host": self.host, "port": self.port},
+                )
+        return db
+    
+    def clear_db(self):
+        # Implement logic to clear the entire VectorDB
```
MichaelClifford:
```suggestion

```

recipes/natural_language_processing/rag/app/manage_vectordb.py
```diff
@@ -0,0 +1,98 @@
...
+                for doc in documents:
+                    collection.add(
+                        ids=[str(uuid.uuid1())],
+                        metadatas=doc.metadata, 
+                        documents=doc.page_content
+                        )
+                print("DB populated")
+            else:
+                print("DB already populated")
+            db = Chroma(client=self.client,
```
MichaelClifford: Do we want to implement the same `.from_documents()` here like below with milvus? https://python.langchain.com/docs/integrations/vectorstores/chroma/

recipes/natural_language_processing/rag/app/rag_app.py
```diff
@@ -1,74 +1,61 @@
...
+vdb_vendor = os.getenv("VECTORDB_VENDOR", "chromadb")
 vdb_host = os.getenv("VECTORDB_HOST", "0.0.0.0")
 vdb_port = os.getenv("VECTORDB_PORT", "8000")
 vdb_name = os.getenv("VECTORDB_NAME", "test_collection")
 
+# Use the following defaults if using milvus db
+# vdb_vendor = os.getenv("VECTORDB_VENDOR", "milvus")
+# vdb_host = os.getenv("VECTORDB_HOST", "127.0.0.1")
+# vdb_port = os.getenv("VECTORDB_PORT", "19530")
+# vdb_name = os.getenv("VECTORDB_NAME", "test_collection")
```
MichaelClifford:
```suggestion

```

recipes/natural_language_processing/rag/app/rag_app.py
```diff
@@ -1,74 +1,61 @@
...
+    return docs
+
 
-vectorDB_client = HttpClient(host=vdb_host,
-                    port=vdb_port,
-                    settings=Settings(allow_reset=True,))
+def create_tmp_file():
+    with open(data.name, mode='wb') as w:
+        w.write(data.getvalue())
+    print(f"{data.name} uploaded")
```
MichaelClifford:
```suggestion

```
remove I don't see this being used anywhere.

recipes/natural_language_processing/rag/app/rag_app.py
```diff
@@ -1,74 +1,61 @@
...
-            metadatas=doc.metadata, 
-            documents=doc.page_content
-            )
-if data == None:
-    print("Empty VectorDB")
+if data != None:
+    with open(data.name, mode='wb') as w:
+        w.write(data.getvalue())
+    print(f"{data.name} uploaded")
+    documents = get_docs(data.name)
```
MichaelClifford: I think this will break on a permissions error if you try to write this file to disk in a container. See how we handled it with `tempfile` in this PR. https://github.com/containers/ai-lab-recipes/pull/330/files

recipes/natural_language_processing/rag/app/rag_app.py
```diff
@@ -1,74 +1,61 @@
...
 else:
-    print("DB already populated")
+    documents = get_docs("../sample-data/fake_meeting.txt")
```
MichaelClifford:
```suggestion

```
I don't think we want to reset `documents` to this hard coded value when `data == None`. I'd just print that the vdb is empty.

vector_dbs/milvus/milvus_standalone_embed.sh
MichaelClifford, line 1: Can we convert this to a makefile or add it to docs instead of using this sh file? Do we actually need anything besides the podman run command here?

MichaelClifford (changes requested):

MichaelClifford:
@Shreyanand You will also need to update `recipes/natural_language_processing/rag/app/Containerfile` to include your new `manage_vectordb.py` file

vector_dbs/milvus/Makefile
```diff
@@ -0,0 +1,55 @@
...
+
+LIB_MILVUS_DIR_MOUNTPATH := $(shell pwd)/volumes/milvus
+
+.PHONY: build
+build:
+	podman build --platform $(PLATFORM) -f Containerfile -t ${IMAGE} .
+
+.PHONY: run
+run:
+	podman run -d \
```
Gregory-Pereira: @Shreyanand I pulled this from your example, but the upstream script runs with `sudo`. I cannot find any reference to why it was put there / if it is actually required. Can you confirm that it ran worked as a standard user?

Gregory-Pereira (commented):
Did not have time to actually run the rag and make sure it works, I just focused on the DB parts and setting up the `Makefile`. I defer to Michael for the actual Rag and DB population bits.

vector_dbs/milvus/Makefile
Gregory-Pereira: Also @Shreyanand please sign your commits. I set DCO to pass just to see if that was blocking RAG tests from running but no dice there. Until then /hold

Shreyanand: @Gregory-Pereira I tried it without sudo and it worked on my local. I am updating the PR with the reviews and DCO checks.

rhatdan:
Needs a rebase.

recipes/natural_language_processing/rag/app/rag_app.py
Shreyanand: If `data == None` then what will the `retriever` object be? The app UI shows an error in the llm chain when it is none. However, once you upload the doc the error goes from the app. Do we want to keep that workflow? Or am I doing something wrong?

MichaelClifford: If there is no data, then the DB would be empty. The retriever would still get instantiated, just to an empty DB.  

I've not seen the error you are referencing.

In any case, I don't think defaulting to a hardcoded file name is the right solution.

Shreyanand: I looked into instantiating the retriever to an empty DB, will update it. 
Agreed on the hardcoded file name, that is not a viable solution.

rhatdan:
Also needs DCO
git commit -a --amend -s
git push --force

.github/workflows/rag.yaml
```diff
@@ -5,16 +5,16 @@ on:
...
+      - '.github/workflows/rag.yaml'
   push:
     branches:
       - main
     paths:
-      - ./recipes/common/Makefile.common
-      - ./recipes/natural_language_processing/rag/**
-      - .github/workflows/rag.yaml
+      - 'recipes/common/Makefile.common'
+      - '/recipes/natural_language_processing/rag/**'
```
Gregory-Pereira:
```suggestion
      - 'recipes/natural_language_processing/rag/**'
```

recipes/natural_language_processing/rag/app/rag_app.py
```diff
@@ -126,4 +98,4 @@ def read_file(file):
     response = chain.invoke(prompt)
     st.chat_message("assistant").markdown(response.content)    
     st.session_state.messages.append({"role": "assistant", "content": response.content})
-    st.rerun()
+    st.rerun()
```
Gregory-Pereira: linting
```suggestion
    st.rerun()

```

Gregory-Pereira (commented):
/lgtm, but 2 tiny nits

Gregory-Pereira (approved):
/lgtm

Gregory-Pereira:
Commit suggestions dont sign with DCO, major bummer. You will have to rebase those commit with signing. Personally I use: 
```
which gamens
gamens: aliased to git commit --amend --no-edit -S -s
```
Alternatively we can just set DCO to pass, but I would prefer stuff stay signed if possible.

vector_dbs/milvus/Makefile
MichaelClifford:
```suggestion
	podman run -it \
```
I'd prefer to see the container output by default

recipes/natural_language_processing/rag/README.md
```diff
@@ -4,7 +4,7 @@ This demo provides a simple recipe to help developers start to build out their o
...
 
-In order for the LLM to interact with our documents, we need them stored and available in such a manner that we can retrieve a small subset of them that are relevant to our query. To do this we employ a Vector Database alongside an embedding model. The embedding model converts our documents into numerical representations, vectors, such that similarity searches can be easily performed. The Vector Database stores these vectors for us and makes them available to the LLM. In this recipe we will use [chromaDB](https://docs.trychroma.com/) as our Vector Database.
+In order for the LLM to interact with our documents, we need them stored and available in such a manner that we can retrieve a small subset of them that are relevant to our query. To do this we employ a Vector Database alongside an embedding model. The embedding model converts our documents into numerical representations, vectors, such that similarity searches can be easily performed. The Vector Database stores these vectors for us and makes them available to the LLM. In this recipe we can use [chromaDB](https://docs.trychroma.com/) or [Milvus](https://milvus.io/) as our Vector Database.
```
MichaelClifford, line 7: @Shreyanand Since the recipe is defined in the ai-lab.yaml file and should only have One vectorDB. I think we should only reference Milvus here and move any mentions of chroma to a chroma readme. Also you should update the ai-lab.yaml file to reference this DB instead.

recipes/natural_language_processing/rag/app/rag_app.py
```diff
@@ -1,91 +1,68 @@
...
 import tempfile
-import uuid
+import streamlit as st
 import os
 
 model_service = os.getenv("MODEL_ENDPOINT","http://0.0.0.0:8001")
 model_service = f"{model_service}/v1"
 chunk_size = os.getenv("CHUNK_SIZE", 150)
 embedding_model = os.getenv("EMBEDDING_MODEL","BAAI/bge-base-en-v1.5")
+vdb_vendor = os.getenv("VECTORDB_VENDOR", "chromadb")
```
MichaelClifford, line 17: lets make milvus the new default.

vector_dbs/README.md
MichaelClifford, line 1: We should follow the same format as the `model_servers/ ` remove this doc, and add a README under each vectorDB directory with all the info to run it.

vector_dbs/milvus/Containerfile
```diff
@@ -0,0 +1,2 @@
+FROM docker.io/milvusdb/milvus:master-20240426-bed6363f
```
MichaelClifford, line 1:
```suggestion
FROM docker.io/milvusdb/milvus:master-20240516-5b27a0cd
```

recipes/natural_language_processing/rag/README.md
MichaelClifford, line 1: You should also add to the readme the step needed to run the application with the new vectorDB.  I think the Env Variables are a little different now.

PR #335, opened 2024-04-25
Intel Habanalabs bootc image

kwozyman:
Add support for Intel Habana Labs, based off RHEL 9.2 drivers for now.

### DIFF
--- a/training/README.md
+++ b/training/README.md
@@ -5,10 +5,11 @@ In order to run accelerated AI workloads, we've prepared [bootc](https://github.
 
 # Makefile targets
 
-| Target     | Description                                   |
-|------------|-----------------------------------------------|
-| amd        | Create bootable container for AMD platform    |
-| nvidia     | Create bootable container for NVidia platform |
+| Target      | Description                                             |
+|-------------|---------------------------------------------------------|
+| amd         | Create bootable container for AMD platform              |
+| nvidia      | Create bootable container for NVidia platform           |
+| intel       | Create bootable container for Intel Habanalabs platform |
 
 # Makefile variables
 
@@ -24,5 +25,5 @@ In order to run accelerated AI workloads, we've prepared [bootc](https://github.
 # Examples
 
 ```
-make amd FROM=registry.redhat.io/rhel9-beta/rhel-bootc:9.4
+make nvidia FROM=registry.redhat.io/rhel9-beta/rhel-bootc:9.4
 ```
--- a/training/Makefile
+++ b/training/Makefile
@@ -4,9 +4,12 @@ help:
 	@echo "Please choose one of the following targets:"
 	@echo "  - amd"
 	@echo "  - nvidia"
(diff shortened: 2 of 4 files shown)


training/README.md
```diff
@@ -5,10 +5,11 @@ In order to run accelerated AI workloads, we've prepared [bootc](https://github.
...
 
-| Target     | Description                                   |
-|------------|-----------------------------------------------|
-| amd        | Create bootable container for AMD platform    |
-| nvidia     | Create bootable container for NVidia platform |
+| Target      | Description                                             |
+|-------------|---------------------------------------------------------|
+| amd         | Create bootable container for AMD platform              |
+| nvidia      | Create bootable container for NVidia platform           |
+| intel-gaudi | Create bootable container for Intel Habanalabs platform |
```
rhatdan: Do you think there will be other intel-nogaudi repos?  I would prefer to stick with just intel.

training/intel-gaudi/Makefile
```diff
@@ -0,0 +1,22 @@
+FROM ?= 
+
+REGISTRY ?= quay.io
+REGISTRY_ORG ?= ai-lab
+IMAGE_NAME ?= intel-gaudi
```
rhatdan: intel-gaudi-bootc
or 
intel-bootc

training/README.md
kwozyman: I'm not sure about the branding to be honest. I know there are some consumer Intel GPUs, but I don't know if those will ever be used for AI or not

training/intel-gaudi/Makefile
kwozyman: I'll say `intel-gaudi-bootc`, just based off my previous comment ^^

training/README.md
rhatdan: I think the same could be said for nvidia and amd GPUs.  I think we should stick to intel, and then if we add some other intel gpus that we want to support but have to do it with other images, then we add intel-OTHERGPU.  With a strong preference for not doing this, and having Intel support all of their GPUs from a RHEL perspective.

kwozyman: Makes sense to me. Should I also rename the directory to just `intel`?

rhatdan: Yes

kwozyman: Done

rhatdan:
LGTM

training/README.md
airlied: intel will have a oneapi variant which is unrelated to this stack, it won't make sense to make a container with both gaudi and oneAPI supported devices (like discrete GPUs).

airlied:
Does this leave out installing the synapseAI stack on purpose? seems like it should at least be mentioned (even if commented out).

rhatdan:
I think we want a default, and over time we will add more variants although I would prefer if we could differentiate at runtime rather then having lots of images.

PR #340, opened 2024-04-25
Add -bootc identifier to training images

rhatdan:
Also set default image in Containerfile and do not set it in Makefile.


Gregory-Pereira (approved):
/lgtm

PR #341, opened 2024-04-25
Add building of instructlab container images

rhatdan:
Currently not sure we need this, but it is worth playing with.

### DIFF
--- a/training/instructlab/Makefile
+++ b/training/instructlab/Makefile
@@ -1,0 +1,23 @@
+default: instructlab
+
+REGISTRY ?= quay.io
+REGISTRY_ORG ?= ai-lab
+IMAGE_TAG ?= latest
+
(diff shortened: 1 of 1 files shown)


stefwalter:
This is working (building) for me.

FWIW, Seems to be needed. There needs to be a way to build all of this together with the images in the amd/nvidia/intel bootable container images.

rhatdan:
Ok removed [WIP]

Just need to figure out how to push them to ai-lab.

stefwalter:
Built nvidia image as: https://quay.io/repository/swalter/instructlab-nvidia

```
$ podman pull quay.io/swalter/instructlab-nvidia:latest
```

stefwalter:
We could use these e2e tests to ensure that these work: https://github.com/instructlab/instructlab/pull/1016

PR #344, opened 2024-04-25
Add clean for models, wastes lots of space

rhatdan:


Gregory-Pereira (approved):
/lgtm
We should probably add the podman-clean make target as seen in the recipes common here as well, because we also have model containers. Happy to do that in another PR though.

PR #345, opened 2024-04-26
add instructlab nvidia base image

sallyom:
let's see how this rolls

### DIFF
--- a/.github/workflows/instructlab_baseimages_build_push.yaml
+++ b/.github/workflows/instructlab_baseimages_build_push.yaml
@@ -77,7 +77,7 @@ jobs:
           registry: ${{ env.REGISTRY }}
 
   build-and-push-image:
-    if: "!contains(github.event.pull_request.labels.*.name, 'hold-tests')"
+    if: "success() && !contains(github.event.pull_request.labels.*.name, 'hold-tests')"
     needs: build-and-push-builder-image
     strategy:
       matrix:
--- a/.github/workflows/instructlab_baseimages_build_push.yaml
+++ b/.github/workflows/instructlab_baseimages_build_push.yaml
@@ -1,0 +1,78 @@
+name: Update quay.io/ai-lab instructlab base images
+
+on:
+  schedule: # schedule the job to run at 12 AM daily
+   - cron: '0 0 * * *'
+
+  pull_request:
+    branches:
+      - main
+    paths:
+      - .github/workflows/instructlab_baseimages_build_push.yaml
+      - training/nvidia
+  push:
+    branches:
+      - main
+    paths:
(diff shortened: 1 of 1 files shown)


.github/workflows/instructlab_baseimages_build_push.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - name: Push image
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        uses: redhat-actions/push-to-registry@v2.8
+        with:
+          image: ${{ steps.build_image.outputs.image }}
+          tags: ${{ steps.build_image.outputs.tags }}
+          registry: ${{ env.REGISTRY }}
+
+  build-and-push-image:
+    if: "!contains(github.event.pull_request.labels.*.name, 'hold-tests')"
```
Gregory-Pereira: We also want to check it passed the first job:
```suggestion
    if: "success() && !contains(github.event.pull_request.labels.*.name, 'hold-tests')"
```

.github/workflows/instructlab_baseimages_build_push.yaml
```diff
@@ -0,0 +1,133 @@
...
+    needs: build-and-push-builder-image
+    strategy:
+      matrix:
+        include:
+          - image_name: instructlab-nvidia
+            label: driver-version=550.54.15 
+            containerfile: training/nvidia/Containerfile
+            context: training/nvidia
+            platforms: linux/amd64
```
Gregory-Pereira, line 89: Feels like we should make this an array

.github/workflows/instructlab_baseimages_build_push.yaml
```diff
@@ -0,0 +1,133 @@
...
+jobs:
+  build-and-push-builder-image:
+    if: "!contains(github.event.pull_request.labels.*.name, 'hold-tests')"
+    strategy:
+      matrix:
+        include:
+          - image_name: nvidia-builder
+            containerfile: training/nvidia/Containerfile.builder
+            context: training/nvidia
+            platforms: linux/amd64
```
Gregory-Pereira, line 35: Feels like this should be an array

Gregory-Pereira:
The 2nd job got stuck waiting to get picked up but the `runs-on` definition was correct. I tried kicking it and we will see wether it does get picked up.  However I did Notice the following log message:
```
Requested labels: ubuntu-latest
Job defined at: containers/ai-lab-recipes/.github/workflows/instructlab_baseimages_build_push.yaml@refs/pull/345/merge
Waiting for a runner to pick up this job...
```
Not sure if the issue is that the job refs include a merge value when this is an open pr ...

Gregory-Pereira (approved):
/lgtm

PR #351, opened 2024-04-27
[Model Upgrade] swapping `whisper-small` with `whisper-large-v3-q5_0`

Gregory-Pereira:
When perusing the [ggerganov whisper.cpp model repo](https://huggingface.co/ggerganov/whisper.cpp/tree/main) we use, I realize that we use this compared to the upstream [OpenAI whisper small repo](https://huggingface.co/openai/whisper-small) there were 2 reasons to use this repo: the `.gguf` format for compatibility with the `llamacpp_python` model_server as well as quantization. However we are not taking advantage of the quantization. This is the only model for which we don't use quantization, so I thought I would fix this.

I selected the `large-v3` model quantized at 5-bits because the footprint of our whisper model was also far below that of our other standard examples of Mistral and Granite. This model size change brings the footprint:

- From:
    - storage: `488 MB`
    - memory: `~1.0 GB`
- To: 
    - storage: `1.08 GB`
    - memory:  `~4.7 GB`

To put this in perspective of our other standard models:

- [mistral-7b-instruct-v0.1.Q4_K_M.gguf](https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/blob/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf)
    - storage: `4.37 GB`
    - memory: `6.87 GB`
- [granite-7b-lab](https://huggingface.co/instructlab/granite-7b-lab)
    - storage: `3.8 GB`
    - No memory usage info available on model card

@MichaelClifford

### DIFF
--- a/.github/workflows/model_image_build_push.yaml
+++ b/.github/workflows/model_image_build_push.yaml
@@ -43,6 +43,10 @@ jobs:
             label: Q4_K_M
             url: https://huggingface.co/instructlab/granite-7b-lab-GGUF/resolve/main/granite-7b-lab-Q4_K_M.gguf
             platforms: linux/amd64,linux/arm64
+          - image_name: whisper-large
+            label: v3_Q5_0
+            url: https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-large-v3-q5_0.bin
+            platform: linux/amd64,linux/arm64
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/ailab-images.md
+++ b/ailab-images.md
@@ -17,3 +17,5 @@
     - [model download link](https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf)
 - quay.io/ai-lab/codellama-7b:latest
     - [model download link](https://huggingface.co/TheBloke/CodeLlama-7B-Instruct-GGUF/resolve/main/codellama-7b-instruct.Q4_K_M.gguf) 
+- quay.io/ai-lab/whisper-large:latest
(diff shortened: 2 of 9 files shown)


rhatdan:
LGTM
But I will let @MichaelClifford decide.

MichaelClifford:
I don't really see a reason to do this at the moment (but happy to be convinced 😄). Have you evaluated the performance of the `ggml-large-v3-q5_0.bin` model? Is it worth it to 2x storage and 5x memory resources for users? There's nothing in our repo that would prevent a user from using the larger more performant model if they want, but to me it doesn't seem like a good practice to use the larger model as default and for testing unless there is a compelling reason.

rhatdan:
I would go with @MichaelClifford thoughts here.

Gregory-Pereira:
I have not, it just seemed odd to me that we use a much smaller model for just this example. Your points are fair, I will convert this to a draft then until I can get some performance metrics to convince you.

Gregory-Pereira:
Regardless @MichaelClifford there is a small quantized version, as well as a small quantized english specific version. Thoughts on using either of those? The quantization should reduce resource utilization.

MichaelClifford:
Small quantized english might be a better option. But one thing to be aware of, when we change the default models here, that needs to be communicated down to the AI Lab team as well. 

> we use a much smaller model for just this example

This is a different kind of model with a different model server so it should not really be compared with the other LLM models. 

> there is a small quantized version, as well as a small quantized english specific version. Thoughts on using either of those?

Its a question of resources vs performance, the quantized models are smaller but generally perform quite a bit worse. We use them essentially out of necessity with the LLM's because un-quantized they are generally too large to be practical. So again it kind of comes down to, do we have a really good reason to choose one of these models over the existing one?

Gregory-Pereira:
Got it, I appreciate you providing more context, seems like I had the conception of "upgrading" the model but it seems more just like a technical choice, and I didn't really understand the technical factors going into model selection. Should I close this then?

PR #352, opened 2024-04-27
[Model Upgrade] upgrade `mistral-7b-instruct-v0.1` to `mistral-7b-instruct-v0.2`

Gregory-Pereira:
[Model `Mistral-7b-instruct-v0.2`](https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.2-GGUF) has been released for a while now, we should upgrade anything using this model. I understand that granite is our default but we still provide it as a working model so the version should get bumped.

@MichaelClifford

### DIFF
--- a/.github/workflows/model_image_build_push.yaml
+++ b/.github/workflows/model_image_build_push.yaml
@@ -33,7 +33,7 @@ jobs:
             platforms: linux/amd64,linux/arm64
           - image_name: mistral-7b-instruct
             label: Q4_K_M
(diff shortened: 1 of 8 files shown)


rhatdan:
This should be handled via renovate.

LGTM

MichaelClifford:
@Gregory-Pereira have you validated that this model works? I was under the impression V2 was not compatible with llamacpp.

MichaelClifford:
**Update**: Tested with our llamacpp model server image. mistral-7b-instruct-v0.2 seems to work well. 😃 Not sure where I got the impression it didn't work.

PR #358, opened 2024-04-28
Fixes for configuring training

rhatdan:


Gregory-Pereira (approved):
Everything looks good, just have not tested it.
/lgtm

Gregory-Pereira:
I would suggest adding a workflow to test them if not for the sizing issues I ran into yesterday.

PR #359, opened 2024-04-28
improper variable definition

cooktheryan:
variable in README is lowercase but in containerfile it is SSHPUBKEY


Gregory-Pereira (approved):
/lgtm
Good catch, `SSHPUBKEY` is used in the `Containerfile` and `Containerfile.nocache`.

sallyom (approved):
LGTM

PR #365, opened 2024-04-29
moving workload to mirrors-repo

Gregory-Pereira:
/cc @lmilbaum + @sallyom , moving this workload to the mirrors-repo.

### DIFF
--- a/.github/workflows/instructlab_baseimages_build_push.yaml
+++ b/.github/workflows/instructlab_baseimages_build_push.yaml
@@ -25,7 +25,7 @@ env:
 
 jobs:
   build-and-push-builder-image:
-    if: "!contains(github.event.pull_request.labels.*.name, 'hold-tests')"
+    if: "!contains(github.event.pull_request.labels.*.name, 'hold-tests') && github.repository == 'containers-mirror/ai-lab-recipes'"
     strategy:
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


lmilbaum (changes requested):
Should this fix the login triggering issue? If not, what does it solve?

Gregory-Pereira:
This was the only difference between this workflow and our model-servers workflow which seems to have no trouble authenticating against the registry and pushing to it. I would assume the registry service is required to interact with the registry, and GitHub understood that and prevented it from running. Although I will need to put some thought into our Podman filesystem usage now it looks like ...

lmilbaum:
> This was the only difference between this workflow and our model-servers workflow which seems to have no trouble authenticating against the registry and pushing to it. I would assume the registry service is required to interact with the registry, and GitHub understood that and prevented it from running. Although I will need to put some thought into our Podman filesystem usage now it looks like ...

I think the service is required for sharing locally built image by multiple jobs. Not sure if that is needed in this workflow.

.github/workflows/instructlab_baseimages_build_push.yaml
```diff
@@ -25,15 +25,15 @@ env:
...
+    if: "!contains(github.event.pull_request.labels.*.name, 'hold-tests') && github.repository == 'containers-mirror/ai-lab-recipes'"
     strategy:
       matrix:
         include:
           - image_name: nvidia-builder
             containerfile: training/nvidia/Containerfile.builder
             context: training/nvidia
             platforms: linux/amd64
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
lmilbaum: Is this change required for the scope of this PR? I actually don't think it is a good idea to use ephemeral images.
Please update the PR description to reflect the change.

lmilbaum (changes requested):

PR #366, opened 2024-04-29
Install rocm-smi package

enriquebelarte:
Adds the rocm-smi package to get info about amdgpu.


kwozyman:
This will significantly decrease the size of this AMD image, as the driver is already included with the kernel.

kwozyman:
Should we also remove the whole EPEL/CRB part? Is that still needed in this case?

rhatdan:
LGTM
Merge when test framework updated.

PR #367, opened 2024-04-29
minor update for `bootc-image-builder` target

YvanY0:
1. `bootc-image-builder` target depends on `bootc`, even the bootc image
exists in the local storage, it still builds a new one. As we add
`--local` in the podman command line, so this commit only check image in
local storage, will not check remote registry.
2. DISK_USER and DISK_GROUP are from #299, they are used for bib option
`--chown`. However, the current variable name may cause
misunderstanding, mislead people into thinking it's a user of operation
system inside disk, so rename them.
Also, add these 2 variables into the README.

Signed-off-by: Yihuang Yu <yihyu@redhat.com>

### DIFF
--- a/recipes/common/Makefile.common
+++ b/recipes/common/Makefile.common
@@ -106,7 +106,12 @@ bootc-run:
 	  $(BOOTC_IMAGE) /sbin/init
 
 .PHONY: bootc-image-builder
-bootc-image-builder: bootc
+bootc-image-builder:
+	@if podman image exists $(BOOTC_IMAGE); then \
+		echo "$(BOOTC_IMAGE) exists in local storage, using it"; \
+	else \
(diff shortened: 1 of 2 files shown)


YvanY0:
At first, I planned to check the remote registry together, but considering that the time to pull the image is about the same as build bootc locally, so I only check local image in this PR. the following is my first version.
```diff
-bootc-image-builder: bootc
+bootc-image-builder:
+	@if ! podman image exists $(BOOTC_IMAGE); then \
+		if ! podman pull $(BOOTC_IMAGE) &> /dev/null; then \
+			$(MAKE) bootc; \
+		fi \
+	fi
```

rhatdan:
LGTM
Waiting to merge until test changes get merged.

Issue #364, opened 2024-04-29
Build InstructLab RHEL AI container images


lmilbaum:
Build the images listed in the following Makefile - https://github.com/containers/ai-lab-recipes/blob/main/training/instructlab/Makefile 
Acceptance Criteria:
1. The images are built and pushed to quay.io/ai-lab once a day
2. They are built with automation. Preferably with GitHub Actions workflow.
3. All images are amd only. Multi-arch is NOT required at the moment.

Gregory-Pereira:
Looking for acceptance criteria from @rhatdan and @sallyom. Many people have asks but no one has be able to clarify what the actual requirements are. From my understanding it would be something along the following lines

- Instructlab images get build and land in quay.io/ai-lab everyday 
- Images are multi-arch

rhatdan:
Instructlab+RHEL AI based images get in.

Gregory-Pereira:
This has been addressed in a number of PRs that landed in main, but this [successful action run](https://github.com/containers-mirror/ai-lab-recipes/actions/runs/8883830153) shows that images are landing in quay.io.

Gregory-Pereira:
This does work however it only tests 2 of the 3 images we want. We also want the intel based one. Therefore I will refactor this to use the make targets, tracked in #384

lmilbaum:
> Instructlab+RHEL AI based images get in.

@rhatdan Would you please map the images you have listed here with the ones listed in the Makefile. Which ones should be built by the end of this week?

lmilbaum:
@Gregory-Pereira Isn't #384 somewhat a duplication of this issue? What would be different in the acceptance criteria from this issue?

Gregory-Pereira:
my understanding was that #364 (this issue) tracks all the acceptance criteria and everything that needs to get done before summit for CI/CD and #384 was just tracking the 1 refactor from building from `buildah` to using make targets

lmilbaum:
@Gregory-Pereira I'm aware that I'm nit picking. For the sake of the board hygiene let's carefully review the acceptance criteria this ticket has and create a new issue for outcomes which are not covered by the existing one.
Implementation details could be shared in this issue. Does that make sense?

Gregory-Pereira:
As far as I can tell, at this point all 3 of our acceptance criteria have been met, but waiting for tomorrow for images to land to say for sure.

- [ ] The images are built and pushed to quay.io/ai-lab once a day  (should be done but confirming tomorrow)
[...]

PR #387, opened 2024-04-29
refactor instructlab training images to use make targets

Gregory-Pereira:
Addresses: #364
changes:
1) add label and arch options to builds
2) extending workflow to `intel-bootc`, `amd-bootc`, and better workflow naming

/cc @sallyom @rhatdan @lmilbaum

### DIFF
--- a/.github/workflows/instructlab_baseimages_build_push.yaml
+++ b/.github/workflows/instructlab_baseimages_build_push.yaml
@@ -28,15 +28,14 @@ env:
   REGISTRY_ORG: ai-lab
 
 jobs:
-  build-and-push-builder-image:
+  nvidia-bootc-builder-image:
     if: "!contains(github.event.pull_request.labels.*.name, 'hold-tests') && github.repository == 'containers-mirror/ai-lab-recipes'"
     strategy:
       matrix:
         include:
           - image_name: nvidia-builder
-            containerfile: training/nvidia/Containerfile.builder
-            context: training/nvidia
-            platforms: linux/amd64
+            context: training/nvidia-bootc
+            archs: amd64
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -56,13 +55,8 @@ jobs:
       
       - name: Build Image
         id: build_image
-        uses: redhat-actions/buildah-build@v2.13
-        with:
-          image: ${{ env.REGISTRY_ORG }}/${{ matrix.image_name }}
-          platforms: ${{ matrix.platforms }}
-          tags: latest
-          containerfiles: ${{ matrix.containerfile }}
-          context: ${{ matrix.context }}
+        run: make dtk ARCH=${{ matrix.archs }}
+        working-directory: ${{ matrix.context }}
 
       - name: Login to Container Registry
         if: github.event_name == 'push' && github.ref == 'refs/heads/main'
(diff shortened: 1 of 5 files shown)


lmilbaum (changes requested):
This PR includes multiple changes which makes it hard to review. How challenging it is to break it into smaller PRs?

Gregory-Pereira:
The `Makefile` changes belong with the workflow changes because they accomodate the different way we are building with labeling the driver version and ARCH vs PLATFORM, I can however remove the slack notification pieces.

FYI I have tested builds and each of these aside from the `nvidia-bootc` we saw earlier can run on standard `ubuntu-latest`. Building with specifying `ARCHS` also works, but once  / if we port this to work for Darwin we will have to refactor it to `PLATFORMS`.

Gregory-Pereira:
I have dropped the slack notifications pieces

.github/workflows/instructlab_baseimages_build_push.yaml
```diff
@@ -28,15 +28,14 @@ env:
...
     if: "!contains(github.event.pull_request.labels.*.name, 'hold-tests') && github.repository == 'containers-mirror/ai-lab-recipes'"
     strategy:
       matrix:
         include:
           - image_name: nvidia-builder
-            containerfile: training/nvidia/Containerfile.builder
-            context: training/nvidia
-            platforms: linux/amd64
+            context: training/nvidia-bootc
+            archs: amd64
```
lmilbaum: Should we use arch instead of archs... to be consistent

.github/workflows/instructlab_baseimages_build_push.yaml
```diff
@@ -80,17 +74,16 @@ jobs:
           tags: ${{ steps.build_image.outputs.tags }}
           registry: ${{ env.REGISTRY }}
 
-  build-and-push-image:
+  nvidia-bootc-image:
     if: "success() && !contains(github.event.pull_request.labels.*.name, 'hold-tests') && github.repository == 'containers-mirror/ai-lab-recipes'"
     needs: build-and-push-builder-image
```
lmilbaum: Job name has changed

.github/workflows/instructlab_baseimages_build_push.yaml
```diff
@@ -102,15 +95,84 @@ jobs:
...
-          labels: |
-            ${{ matrix.label }}
-          tags: latest
-          containerfiles: ${{ matrix.containerfile }}
-          context: ${{ matrix.context }}
+          image: ${{ steps.build_image.outputs.image }}
+          tags: ${{ steps.build_image.outputs.tags }}
+          registry: ${{ env.REGISTRY }}
+
+  intel-bootc-image:
```
lmilbaum, line 117: This and the next job looks very similar to the first one. Is it possible to consolidate that with the matrix section?

lmilbaum (changes requested):

.github/workflows/instructlab_baseimages_build_push.yaml
Gregory-Pereira: This name change was intentional, but if you dont like it I can revert it. I thought since there are 4 images that are considered "instructlab base" images, that we should not have generic workflow names like `build-and-push-image`. This way you know what image the step is actually building. I understand that I will have to delete the old workflow from the mirror repo. Thoughts?

.github/workflows/instructlab_baseimages_build_push.yaml
Gregory-Pereira: We could consolidate jobs 3 and 4, this change is coming

.github/workflows/instructlab_baseimages_build_push.yaml
Gregory-Pereira: The idea was that `archs` was an array, like: `archs: "amd64,arm64"`. I know we only need `amd64` for now but I thought `archs` array made it extendable. However in retrospect we would need a refactor anyway for `platform` so I accept this suggestion and will make it `arch`.

.github/workflows/instructlab_baseimages_build_push.yaml
lmilbaum: Sorry for providing a confusing comment. This job is depended on another job which is listed in the needs key. The other job name was renamed but this one was not changed. I think it should be changed as well.

Gregory-Pereira: Thank you for the clarification. This has been addressed now.

lmilbaum (approved):
LGTM

PR #389, opened 2024-04-29
download models from models dir

Gregory-Pereira:
Identified via: https://github.com/containers/ai-lab-recipes/actions/runs/8885812868/job/24397987802?pr=388#step:7:1
Models should be downloaded from models dir.


lmilbaum (approved):

lmilbaum:
That is the problem when code has changed while the CI filtered some workflows which could have shown that the code changes broke CI

PR #394, opened 2024-04-30
reworking workflow names for visibility from actions tab

Gregory-Pereira:
When looking from the actions tab, job is named `build-and-push-image`. This causes me to constantly lose my context and increase difficulty debugging. I want job names to understand which workflow and job is running by the job name itself. Not sure if there are better naming conventions, but WDYT @lmilbaum


rhatdan:
LGTM

lmilbaum (approved):
LGTM

PR #397, opened 2024-04-30
Improve summary chunking

MichaelClifford:
Document chunking in `summarizer.py` was still failing with some models on some texts. (There were some model vocabularies that caused the model server to fail to return the detokenized data if specific token sequences were split into separate chunks by the chunking process).

To address this issue, we've implemented a new approach to chunking. Instead of chunking the raw tokens, we will chunk the text, like in our earlier implementation, and rely on `extras/tokenize/count` to confirm that the text chunk is within the token limit, if not, we split the text chunk in two. 

I've manually evaluated this with the problematic and non-problemtatic models. It appears to work as well as before, and avoids any of the detokenization crashes we were experiencing.

### DIFF
--- a/recipes/natural_language_processing/summarizer/app/summarizer.py
+++ b/recipes/natural_language_processing/summarizer/app/summarizer.py
@@ -3,6 +3,7 @@ from langchain_openai import ChatOpenAI
 from langchain.prompts import PromptTemplate
 from langchain_community.callbacks import StreamlitCallbackHandler
 from langchain_community.document_loaders import PyMuPDFLoader
+from langchain_text_splitters import RecursiveCharacterTextSplitter
 from rouge_score import rouge_scorer
 import streamlit as st
 import tempfile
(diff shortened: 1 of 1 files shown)


recipes/natural_language_processing/summarizer/app/summarizer.py
```diff
@@ -34,22 +35,32 @@ def checking_model_service():
...
-    for i in range(num_chunks):
-        chunk = tokens[:chunk_size]
-        chunk = requests.post(f"{model_service[:-2]}extras/detokenize/",
-                  json={"tokens":chunk}).content
-        chunk = json.loads(chunk)["text"]
-        chunks.append(chunk)
-        tokens = tokens[chunk_size:]
+    text_splitter = RecursiveCharacterTextSplitter(
+        chunk_size=3048,
+        chunk_overlap=0,
```
Gregory-Pereira, line 48: I feel like these sorts of applications usually have overlap for context in the summarization. Could you elaborate on why this wasn't chosen?

recipes/natural_language_processing/summarizer/app/summarizer.py
```diff
@@ -34,22 +35,32 @@ def checking_model_service():
...
+        is_separator_regex=False
+        )
+    
+    text_chunks = text_splitter.create_documents([text])
+    for chunk in text_chunks:
+        chunk = chunk.page_content
+        count = requests.post(f"{model_service[:-2]}extras/tokenize/count",
+                  json={"input":chunk}).content
+        count = json.loads(count)["count"]
+        if count >= 2048:
```
Gregory-Pereira, line 59: Is this meant to be `2048`, or is the [above character count](https://github.com/containers/ai-lab-recipes/pull/397/files#diff-2922bab9dc3458bfc5476bde5cb41e9b43f28a4c2287db54bea7a4403aafa8a6R47) meant to be `2048` instead of `3048`? If not why different values?

Gregory-Pereira (commented):
/lgtm, but I had some questions. I also haven't tested this, but if you are confident it works we can ignore that concern

recipes/natural_language_processing/summarizer/app/summarizer.py
MichaelClifford: I wanted to minimize the number of tokens processed since this application can take a while to run with longer docs. Not including overlap in the chunks seemed like a reasonable place to limit total token count. could be adjusted in the future if needed.

recipes/natural_language_processing/summarizer/app/summarizer.py
MichaelClifford: Its a bit of a hand tuned process here. But, 2048 is the default context length of the llamacpp server we are using. So any token count greater than 2048  will generate a `500` error from the model server. (its actually a bit smaller than that when you include the prompt template and the output text size). So we want to split the chunk further if it is greater than 2048 tokens. 

The 3048 is the max number of words, not tokens, that can be in a chunk. My initial attempt at setting chunk_size was to set it to 2048 as well. However, I found the `RecursiveCharacterTextSplitter()` to be rather conservative, causing pretty small chunk sizes (much smaller than 2048), meaning more chunks and more loops of the summary process (and more time). So I increased this number until I had the same number of chunks per test document as I did with the other method while not breaking.   

Future iterations could replace these hardcoded numbers with values from the llamacpp server, to ensure the context lengths here and those used by the model server always match.

Gregory-Pereira: Understood, these seem reasonable. I was under the misunderstanding they were unintentionally different.

Gregory-Pereira (approved):
/LGTM

PR #398, opened 2024-04-30
model container builder

kwozyman:


training/model/Makefile
```diff
@@ -0,0 +1,22 @@
+FROM ?= 
+
+REGISTRY ?= quay.io
+REGISTRY_ORG ?= ai-lab
+IMAGE_NAME ?= granite-model
```
Gregory-Pereira: Required change: In `quay.io/ai-lab` the granite model image is: [granite-7b-lab](https://quay.io/repository/ai-lab/granite-7b-lab)
```suggestion
IMAGE_NAME ?= granite-7b-lab
```

training/model/Makefile
```diff
@@ -0,0 +1,22 @@
...
+IMAGE_TAG ?= latest
+
+CONTAINER_TOOL ?= podman
+CONTAINER_TOOL_EXTRA_ARGS ?=
+
+MODEL_REPO ?= ibm/granite-7b-base
+MODEL_PATH ?= /usr/share/ai-model
+
```
Gregory-Pereira, line 13: Optional change: Can we add an `ARCH` variable and pass it to our build command? I know we only want to support `amd64` ATM, however a lot of our team develop on Apple silicon which is `arm64`. This helps me to test / build these `amd64` images by just adding the `ARCH` override.

Gregory-Pereira (commented):
/lgtm, but 1 required change and 1 optional change

Gregory-Pereira (approved):
Required change fixed, marking this as approved. If you want to include the optional change feel free.

training/model/Makefile
kwozyman: I don't think we need an extra var for that. You can just use `CONTAINER_TOOL_EXTRA_ARGS="--arch arm64"` and that should work everywhere, not only in the model.

PR #402, opened 2024-04-30
Add hl-smi binary from firmware-tools

enriquebelarte:
Adding hl-smi to image so hardware can be queried


Gregory-Pereira (approved):
/lgtm, just know I have no bootc knowledge but if this works go for it

kwozyman:
It builds correctly with Stream and RHEL and I think @enriquebelarte also tested on bare metal.

PR #405, opened 2024-04-30
create basic instructlab testing framework

Gregory-Pereira:
changes:

1. Create base `Instructlab_testing_framework` workflow. Currently provisioning t3.micro with 25 gigs of storage, but will up it when I have confirmed provisioning itself works. It should only bootstrap the ENV and then setup python and tear itself down and communicate results to slack -- very basic to start.
2. Point `instructlab` base workflow to produce the images in `ghrc.io/containers` rather than `quay.io/ai-lab`. This change will be leveraged in the second pass at the `Instructlab_testing_framework`, which will run the actual e2e tests, and if successful, copy those images into `quay.io/ai-lab`

/cc @lmilbaum @sallyom

### DIFF
--- a/.github/workflows/instructlab.yaml
+++ b/.github/workflows/instructlab.yaml
@@ -24,8 +24,8 @@ concurrency:
   cancel-in-progress: false
 
 env:
-  REGISTRY: quay.io
-  REGISTRY_ORG: ai-lab
+  REGISTRY: ghcr.io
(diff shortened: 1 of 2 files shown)


Gregory-Pereira:
Tests are not running because they are running in my fork rather than in in `containers/ai-lab-recipes`, going to reopen this pr in an upstream branch.

PR #408, opened 2024-04-30
add vllm server container, ability to start the server, ilab generate

cdoern:
we now have the ability to do a top level `serve` in the ilab-wrapper which will start a vllm server, and run `generate` which will connect to the ENDPOINT_URL specified

### DIFF
--- a/training/vllm/Containerfile
+++ b/training/vllm/Containerfile
@@ -1,0 +1,5 @@
+FROM quay.io/wxpe/tgis-vllm:release.4e3ff78
+
+USER root
+RUN ln -s /usr/lib64/libcuda.so.1 /usr/lib64/libcuda.so
+COPY mixtral.jinja .
--- a/training/Makefile
+++ b/training/Makefile
@@ -5,14 +5,17 @@ help:
 	@echo "  - amd"
 	@echo "  - nvidia"
 	@echo "  - intel"
+	@echo "	 - vllm"
 
-.PHONY: amd nvidia intel
+.PHONY: amd nvidia intel vllm
 amd:
 	make -C amd-bootc/ bootc
 nvidia:
 	make -C nvidia-bootc/ dtk bootc
 intel:
 	make -C intel-bootc/ bootc
+vllm:
+	make -C vllm/ image
 
 .PHONY:
 instruct-nvidia:
@@ -27,3 +30,4 @@ instruct: instruct-nvidia instruct-amd
 
 clean:
 	rm -rf build
+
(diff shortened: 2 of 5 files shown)


cdoern:
cc @n1hility how does this look for vllm. I am working on testing locally, but just making sure the general structure makes sense. This deals with the vllm serve, and the ilab generate.

training/vllm/Containerfile
```diff
@@ -0,0 +1,6 @@
+FROM quay.io/wxpe/tgis-vllm:release.4e3ff78
+
+USER root
+RUN ln -s /usr/lib64/libcuda.so.1 /usr/lib64/libcuda.so
+COPY mixtral.jinja .
+COPY ilab_to_sdg.py .
```
jaideepr97: dont need this one here i think

training/vllm/Containerfile
```diff
@@ -0,0 +1,6 @@
+FROM quay.io/wxpe/tgis-vllm:release.4e3ff78
+
+USER root
+RUN ln -s /usr/lib64/libcuda.so.1 /usr/lib64/libcuda.so
+COPY mixtral.jinja .
```
rhatdan: I would prefer the Model to be on the host (bootc container) rather then inside of one particular container this would give us the ability to share it with other containers potentially.

rhatdan:
Are we going to want the vllm container embeded into the nvidia, amd and intel containers?

training/ilab-wrapper/ilab
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 
 # Template values replaced by container build
+ENDPOINT_URL="__REPLACE_ENDPOINT_URL__"
+DATE="__REPLACE_DATE__"
+TAXONOMY_PATH="__REPLACE_TAXONOMY_PATH__"
```
n1hility: I think we should put these changes in another name ilab-full for example, just so that everyone still functions until we finish

n1hility: sorry to be clear i mean instead of against ilab another file, that way the image is still consumable then once done we can flip it

training/vllm/Containerfile
cdoern: removing!

training/vllm/Containerfile
cdoern: this is just a jinja template, probably just needed for vllm as this is used to serve the model.

n1hility:
> Are we going to want the vllm container embeded into the nvidia, amd and intel containers?

Yes although starting with just nvidia for now

PR #409, opened 2024-04-30
Pull instructlab images out of local storage

rhatdan:
Also create a Makefile.common in training so that the different hardware vendors Makefiles can share options.

### DIFF
--- a/.github/workflows/training_bootc.yaml
+++ b/.github/workflows/training_bootc.yaml
@@ -107,7 +107,7 @@ jobs:
       
       - name: Build Image
         id: build_image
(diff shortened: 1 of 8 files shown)


n1hility:
LGTM

Gregory-Pereira:
If you can @rhatdan include the workflow changes when possible.

rhatdan:
@Gregory-Pereira what did I change in workflows?

Gregory-Pereira:
The workflows used those make targets to build. I added a commit updating the workflows to use the new `bootc` make target instead of the `image` target it was previously. Were close to testing about 65% makefile coverage in our tests, odds are whenever make targets change it will affect testing infra.

rhatdan:
Ok. sorry.

This patch is broken.  Will be opening another.

PR #414, opened 2024-05-01
Add containerfile for deepspeed training container

jaideepr97:


training/deepspeed/Containerfile
```diff
@@ -0,0 +1,17 @@
+# Containerfile for running deepspeed training 
+FROM nvidia/cuda:12.1.1-cudnn8-devel-ubi9
+
+RUN dnf install -y python python-devel git
+RUN python -m ensurepip --upgrade
+RUN pip3 install torch==2.1.2 --index-url https://download.pytorch.org/whl/cu121
```
Gregory-Pereira: I feel like all of this could be a `requirements.txt` that you add to the container, is there a reason to do it this way instead?

training/deepspeed/Makefile
```diff
@@ -0,0 +1,18 @@
...
+KERNEL_VERSION ?= 
+
+default: image
+
+.PHONY: image
+image:
+	"${CONTAINER_TOOL}" build \
+			$(ARCH:%=--platform linux/%) \
+			--file Containerfile \
+			--tag "${REGISTRY}/${REGISTRY_ORG}/${IMAGE_NAME}:${IMAGE_TAG}" \
```
Gregory-Pereira, line 15: Are the `DRIVER_VERSION` and `KERNEL_VERSION` meant to be passed as extra build-args? They seem like they are currently unused ...

Gregory-Pereira (commented):
Linting + 2 other comments, looks fine to me though. That being said I know next to nothign about the ilab / training bits so maybe wait for someone a bit more senior to merge unless your sure

training/deepspeed/Containerfile
jaideepr97: updated, thanks!

training/deepspeed/Makefile
jaideepr97: removed, thanks!

n1hility:
LGTM

PR #410, opened 2024-04-30
Add growfs service to chatbot/bootc Containerfiles

sallyom:
@dwalsh  @cgwalters testing this now, added colin as author bc i copied his files from bootc-org/examples 

Tested with an `ami` with (and also with `podman-bootc`):
```
$ cd recipes/natural_language_processing/chatbot
$ make BOOTC_IMAGE=quay.io/sallyom/centos-bootc:chatbot ARCH=x86_64 CONTAINERFILE=Containerfile.nocache bootc
$ aws s3 cp ~/Desktop/bootc-disk-images/image/disk.raw s3://BUCKET_NAME
$ aws ec2 import-snapshot  --description "centos-bootc-chatbot" --disk-container file://ami-container.json
$ aws ec2 describe-import-snapshot-tasks --filters Name=task-state,Values=active

# create image (ami) from snapshot in console

$ cat ami-container.json 
{
   "Description": "centos-bootc-chatbot",
   "Format": "raw",
   "Url": "s3://BUCKET_NAME/disk.raw"
}
```

PR #415, opened 2024-05-01
Add cloud-init build feature

rhatdan:
Improve help and README to describe new functionality.

### DIFF
--- a/training/Makefile
+++ b/training/Makefile
@@ -1,15 +1,71 @@
 default: help
 
 help:
-	@echo "Please choose one of the following targets:"
-	@echo "  - amd"
-	@echo "  - nvidia"
-	@echo "  - intel"
+	@echo "To build a bootable container image you first need to create instructlab container images for a particular vendor "
+	@echo
+	@echo "  - make instruct-amd"
+	@echo "  - make instruct-intel"
+	@echo "  - make instruct-nvidia"
+	@echo "  - make instruct-vllm"
+	@echo
+	@echo "Once instruct images are created, creat bootc container images"
+	@echo
+	@echo "  - make bootc-amd"
+	@echo "  - make bootc-intel"
+	@echo "  - make bootc-nvidia"
+	@echo "  - make bootc-vllm"
+	@echo
+	@echo "If these images are going to be used on a cloud, you might want to add cloud-init."
+	@echo
+	@echo "  - make cloud-amd"
+	@echo "  - make cloud-intel"
+	@echo "  - make cloud-nvidia"
+	@echo "  - make cloud-vllm"
 
-.PHONY: amd nvidia intel
+#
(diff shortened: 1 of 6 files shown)


rhatdan:
@n1hility @cgwalters PTAL

training/Makefile
```diff
@@ -1,15 +1,71 @@
...
+	make -C nvidia-bootc/ dtk bootc
+vllm:
+	make -C vllm/ image
+
+#
+# Make Bootc container images preinstalled with cloud-init
+#
+.PHONY:
+cloud-amd:
+	make VENDOR=amd -C cloud
```
cgwalters, line 57: Is there a missing `git add`?  I don't see this directory in the PR

training/README.md
```diff
@@ -57,6 +65,15 @@ Of course, the other Makefile variables are still available, so the following is
 make nvidia REGISTRY=myregistry.com REGISTRY_ORG=ai-training IMAGE_NAME=nvidia IMAGE_TAG=v1 FROM=registry.redhat.io/rhel9/rhel-bootc:9.4
 ```
 
+# How to build Cloud ready images
+
+Bootc container images can be installed on physical machines, virtual machines and in the cloud.  Often it is useful to add the cloud-init package when running the operarting systems in the cloud.
```
cgwalters, line 70: I think we can shortly redirect to some upstream docs on this btw.

But the larger problem is that cloud-init isn't currently working for the growfs case so we should do https://github.com/containers/ai-lab-recipes/pull/410 first.

training/README.md
```diff
@@ -1,15 +1,21 @@
...
-| intel       | Create bootable container for Intel Habanalabs platform |
+| Target          | Description                                                         |
+|-----------------|---------------------------------------------------------------------|
+| instruct-amd    | Create instruct lab image for bootable container for AMD platform   |
+| instruct-intel  | Create instruct lab image for bootable container for Intel platform |
+| instruct-nvidia | Create instruct lab image for bootable container for Nvidia platform|
+| amd             | Create bootable container for AMD platform                          |
+| intel           | Create bootable container for Intel Habanalabs platform             |
+| nvidia          | Create bootable container for NVidia platform                       |
+| cloud-amd       | Add cloud-init to bootable container for AMD platform               |
```
cgwalters, line 16: It feels like it'd be nicer to have this as flag, something like

```
make amd INCLUDES=cloud-init
```

Would generalize more obviously to including other snippets in the future.

training/Makefile
rhatdan: Yup added.

training/README.md
rhatdan: Sounds good lets merge and then I can work on this. Would involve more complicated Makefile gymnastics.

PR #419, opened 2024-05-01
Fix Nvidia target name

kwozyman:


Gregory-Pereira (changes requested):
Seems that all the other make targets have been moved back to `image` rather than `bootc`. I do not think this change is correct, but please let me know if I am misunderstanding something

kwozyman:
Hmm, I see in the main training/Makefile we still call "bootc" for all platforms

Gregory-Pereira:
Got it, I wasn't checking the Makefile for the traning dir, it seems that you are right that [makefile calls it bootc](https://github.com/containers/ai-lab-recipes/blob/main/training/Makefile#L44-L48). However, all of the other Makefiles ([intel-bootc](https://github.com/containers/ai-lab-recipes/blob/main/training/intel-bootc/Makefile#L16-L17), [nvidia-bootc](https://github.com/containers/ai-lab-recipes/blob/main/training/nvidia-bootc/Makefile#L41-L42), [amd-bootc](https://github.com/containers/ai-lab-recipes/blob/main/training/amd-bootc/Makefile#L13-L14) `Makefiles` all use `image` instead of `bootc`. I would love some clarification from @rhatdan on what we want this to be.

training/nvidia-bootc/Makefile
```diff
@@ -39,7 +39,7 @@ dtk:
 		.
 
 .PHONY: image
```
rhatdan, line 41: This should be bootc as well, Well should have nothing I believe.

rhatdan:
Everything should be bootc, not image.

rhatdan:
This looks like it is fixed in main.

PR #425, opened 2024-05-01
Use ARCH variable, so we fail early on aarch64

ericcurtin:
When using this on alternate platform like aarch64, at least it will cause the make script to fail early when the aarch64 file doesn't exist, rather than incorrectly downloading the amd64 version. aarch64 chrome may get packaged in future also.

### DIFF
--- a/recipes/common/Makefile.common
+++ b/recipes/common/Makefile.common
@@ -51,7 +51,7 @@ ifeq ($(OS),Darwin) # This structure may vary if we upgrade chromedriver, see in
 	CHROME_DOWNLOAD_PATH := googlechrome.dmg
 else ifeq ($(OS),Linux)
 	CHROMEDRIVER_DOWNLOAD_PATH := chromedriver_linux64.zip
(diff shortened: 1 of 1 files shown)


Gregory-Pereira (approved):
Lets go with this for now, but maybe you could make an issue around this experience for the future. We should consider upgrading crome + driver versions, finding a more official source or something else. There is also alternate ways of naming the arch to consider like: `x86_64` vs `amd64` and `aarch64` vs `arm64` ...

ericcurtin:
True. aarch64/arm64 .deb Chrome doesn't exist, but we might get lucky and see it sometime, I think it's the last platform left they don't build for pretty much... non-ChromeOS aarch64 Linux (.deb/.rpm).

ericcurtin:
I think this handling was ok earlier in the script:

```
ARCH := $(shell uname -m)
ifeq ($(ARCH),x86_64)
	ARCH := amd64
endif
```

that's the best you can do in these situations I think, just pick one.

rhatdan:
LGTM

PR #426, opened 2024-05-01
Add object detection model server readme

MichaelClifford:
Added a readme for object detection model server. Updated the make file so that is uses the newer model download script.


model_servers/object_detection_python/README.md
```diff
@@ -0,0 +1,52 @@
...
+## Deploy Model Server
+
+The model server relies on a volume mount to the localhost to access the model files. It also employs environment variables to dictate the model used and where its served. You can start your model server using the following `make` command from the [`model_servers/object_detection_python`](../../../model_servers/object_detection_python) directory, which will be set with reasonable defaults:
+
+```bash
+make run
+```
+or 
+
+```bash
```
Gregory-Pereira, line 45: might want to include the cd to models dir before the run. This is because you cannot traverse upward on a volume mount

Gregory-Pereira (approved):
1 optional suggestion but everything looks good. Ill finish up the rest in my PR later today

model_servers/object_detection_python/README.md
MichaelClifford: I usually just use an absolute path for volume mounts

rhatdan:
LGTM

PR #427, opened 2024-05-01
Use OCI Transport for embeded images to save time

rhatdan:
No reason to use containers/storage for instructlab or vllm since we are only building for embedding within a bootc image. By storing directly in OCI, we can save many minutes and lots of disk size.


n1hility:
LGTM

lmilbaum (approved):
LGTM. Would be useful if the work @Gregory-Pereira and I are doing can validate this change.

sallyom:
LGTM, also - as long as it works (I haven't tested this)!

PR #428, opened 2024-05-02
bootc e2e workflow - build temp image shim with ssh keys around bootc 

Gregory-Pereira:
The previous PRs have setup our ENV and given us access to the bootc installed instance, now we want to run the e2e tests

### DIFF
--- a/.github/workflows/training-e2e.yaml
+++ b/.github/workflows/training-e2e.yaml
@@ -90,6 +90,10 @@ jobs:
         env:
           ANSIBLE_HOST_KEY_CHECKING: false
 
+      - name: log some networking information
+        run: |
+          ifconfig
+
       - name: Destroy Test Environment
         id: down
         if: always()
--- a/.github/workflows/training-e2e.yaml
+++ b/.github/workflows/training-e2e.yaml
@@ -15,9 +15,9 @@ on:
 
 env:
   TF_VAR_aws_region: "eu-west-2"
-  TF_VAR_aws_ami_owners: '["125523088429"]'
-  TF_VAR_aws_ami_name: '["Fedora-Cloud-Base-*"]'
-  TF_VAR_aws_volume_size: 128
+  TF_VAR_aws_ami_owners: '["309956199498"]'
+  TF_VAR_aws_ami_name: '["*RHEL-9.4*"]'
+  TF_VAR_aws_volume_size: 500
   TF_VAR_aws_access_key: ${{ secrets.AWS_ACCESS_KEY_ID }}
   TF_VAR_aws_secret_key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
 
@@ -74,6 +74,7 @@ jobs:
         run: |
           echo "id=$(terraform output id | xargs)" >> $GITHUB_OUTPUT
           echo "url=$(terraform output host | xargs)" >> $GITHUB_OUTPUT
+          echo "ssh_public_key=$(terraform output ssh_public_key | xargs)" >> $GITHUB_OUTPUT
           echo "pem_filename=$(terraform output pem_filename | xargs)" >> $GITHUB_OUTPUT
         working-directory: terraform-test-environment-module
 
@@ -83,16 +84,23 @@ jobs:
 
       - name: Provision
         run: |
+          ssh_public_key=$(printf '%s\n' "${{ steps.terraform-output.outputs.ssh_public_key }}" | sed -e 's/[\/&]/\\&/g')
           ansible-playbook ./main/training/provision/playbook.yml \
+            -vvv \
             -i terraform-test-environment-module/hosts.ini \
             --private-key=terraform-test-environment-module/${{ steps.terraform-output.outputs.pem_filename }} \
-            --extra-vars "image_name=${{ matrix.image_name }}" 
+            --extra-vars "image_name=${{ matrix.image_name }}" \
+            --extra-vars "ssh_public_key=$ssh_public_key" \
+            --extra-vars "registry_user=${{ secrets.REGISTRY_USER }}" \
+            --extra-vars "registry_password=${{ secrets.REGISTRY_PASSWORD }}"
         env:
           ANSIBLE_HOST_KEY_CHECKING: false
 
-      - name: log some networking information
+      - name: run the e2e tests
         run: |
-          ifconfig
+          ssh -i terraform-test-environment-module/${{ steps.terraform-output.outputs.pem_filename }} \
+          ${{ steps.terraform-output.outputs.url }} \
+          curl -sLO /tmp https://raw.githubusercontent.com/stefwalter/instructlab/container-testing/scripts/basic-workflow-tests.sh
 
       - name: Destroy Test Environment
         id: down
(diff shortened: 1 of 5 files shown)


.github/workflows/training-e2e.yaml
```diff
@@ -17,7 +17,7 @@ env:
   TF_VAR_aws_region: "eu-west-2"
   TF_VAR_aws_ami_owners: '["125523088429"]'
   TF_VAR_aws_ami_name: '["Fedora-Cloud-Base-*"]'
-  TF_VAR_aws_volume_size: 128
+  TF_VAR_aws_volume_size: 500
```
cgwalters, line 20: Does this one also possibly need changing? https://github.com/containers/ai-lab-recipes/blob/1e39174d03409c2874c7a02d36182654d0d6f404/.github/workflows/testing_framework.yaml#L17

training/provision/playbook.yml
```diff
@@ -15,16 +15,41 @@
...
       name: podman
       state: present
 
+  - name: Login to default registry and create ${XDG_RUNTIME_DIR}/containers/auth.json
+    containers.podman.podman_login:
+      username: "{{ registry_user }}"
+      password: "{{ registry_password }}"
+      registry: quay.io
+
+  - name: Building an image with ssh key
```
cgwalters: Wait...unless I am misunderstanding the ansible abstractions here, this isn't doing a build at all.  This is just running some commands in the container and throwing away the result.

cgwalters (changes requested):

training/provision/playbook.yml
cgwalters: IOW, the more desired flow here would be to build a derived image, push it to a registry (so the host can do in-place updates, if we care to test that) and then also use the bootc install flow.

Alternatively, it works to do a local build for testing and install that, but it obviously disables in-place updates, in case we care about that.

training/provision/playbook.yml
```diff
@@ -15,16 +15,41 @@
...
+      image: "quay.io/ai-lab/{{ image_name }}:latest"
+      command:
+      - mkdir /usr/etc-system
+      - test -n "{{ ssh_public_key }}"
+      - echo 'AuthorizedKeysFile /usr/etc-system/%u.keys' >> /etc/ssh/sshd_config.d/30-auth-system.conf
+      - "echo {{ ssh_public_key }} > /usr/etc-system/fedora.keys && chmod 0600 /usr/etc-system/fedora.keys"
+      state: present
+      recreate: true
+
   - name: Bootc install
```
cgwalters: Per discussion in chat this will not work with any centos/rhel derived container images targeting a Fedora cloud instance which uses btrfs.  See https://github.com/containers/bootc/issues/192

training/provision/playbook.yml
lmilbaum: Thanks @cgwalters for catching it.  The module was replaced with the correct one now.
Regarding the in-place updates, that is definitely needed. Lets file a GH issue in this repo and it will be handled in a followup PR.

training/provision/playbook.yml
```diff
@@ -15,16 +15,51 @@
...
+      authfile: /etc/containers/auth.json
+
+  - name: Build Derived Image
+    async: 1000
+    poll: 0
+    register: build_result
+    containers.podman.podman_image:
+      name: "quay.io/ai-lab/{{ image_name }}:latest"
+      build:
+        target: localhost/derivied_image:latest
```
cgwalters: Typo "derived"

training/provision/playbook.yml
```diff
@@ -37,7 +72,9 @@
       privileged: yes
       pid_mode: host
       command: "bootc install to-filesystem --karg=console=ttyS0,115200n8 --replace=alongside /target"
+    become: true
```
cgwalters: I don't believe this is necessary as there's a toplevel `become: true` on the whole playbook.

training/provision/playbook.yml
```diff
@@ -15,16 +15,51 @@
   - name: Gather facts for first time
     ansible.builtin.setup:
 
-  - name: Check Podman Present
-    ansible.builtin.package:
-      name: podman
+  - name: Required packages
+    ansible.builtin.dnf:
+      name:
+      - https://s3.eu-west-2.amazonaws.com/amazon-ssm-eu-west-2/latest/linux_amd64/amazon-ssm-agent.rpm
```
cgwalters, line 21: I'm curious why...this beast has deep hooks into everything, it's Amazon wanting to own systems management.

training/provision/playbook.yml
```diff
@@ -15,16 +15,51 @@
...
+      - podman
       state: present
+      disable_gpg_check: true
 
-  - name: Bootc install
+  - name: Derived Image Containerfile
+    ansible.builtin.template:
+      src: ./templates/Containerfile.j2
+      dest: /tmp/Containerfile
```
cgwalters, line 29: Minor nit but, predictable filenames in `/tmp` are a CWE: https://capec.mitre.org/data/definitions/149.html

There's a [tempfile module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/tempfile_module.html).

Not a big deal, but just noting.

training/provision/playbook.yml
cgwalters: But um...actually looking at the ansible docs for this module I see
https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_image_module.html#parameter-build/target
"Specify the target build stage to build." which doesn't sound like what's desired here...

I think what you want is to use what you have in `target` for `name` actually. 

Listen so...I like Ansible, I think it's a powerful swiss army knife that absolutely has its uses.

But this specific code, it doesn't seem like the ansible abstractions are buying us anything over plain old basic shell script executing `podman build` + `podman run`...

training/provision/templates/Containerfile.j2
```diff
@@ -0,0 +1,9 @@
+FROM quay.io/ai-lab/{{ image_name }}:latest
```
cgwalters, line 1: `podman build` already has `--from` to template this without writing a new containerfile btw.

training/provision/templates/Containerfile.j2
```diff
@@ -0,0 +1,9 @@
...
+
+USER root
+
+RUN mkdir /usr/etc-system && \
+    chown -R root:root /usr/etc-system && \
+    echo 'AuthorizedKeysFile /usr/etc-system/root.keys' >> /etc/ssh/sshd_config.d/30-auth-system.conf && \
+    echo {{ ssh_public_key }} > /usr/etc-system/root.keys && \
```
cgwalters, line 8: And this one I think is better as a build argument instead of jinja templating, which can be passed already via `--build-arg`.

training/provision/playbook.yml
lmilbaum: I use it for connecting to the instance for debugging purposes. It will be removed as soon as the workflow starts working

training/provision/playbook.yml
```diff
@@ -15,29 +15,42 @@
...
+
+  - name: Login to default registry
+    containers.podman.podman_login:
+      username: "{{ registry_user }}"
+      password: "{{ registry_password }}"
+      registry: quay.io
+      authfile: /etc/containers/auth.json
```
cgwalters, line 36: Yeah, we really want https://github.com/containers/image/pull/1746 ...will see if I or someone else can get back to it.

PR #429, opened 2024-05-02
created cloud init workflow and documentation

Gregory-Pereira:
/cc @lmilbaum I need these to build the images. I cannot access the container even after ignoring failures which cause any steps in the Ansible playbook to fail, but the github action thinks its succeeded, so it just ignores those sections.


Gregory-Pereira:
moved to #430

PR #430, opened 2024-05-02
created cloud init workflow and documentation

Gregory-Pereira:
Dupe of #429 but moved to upstream branch to enable testing

### DIFF
--- a/.github/workflows/cloud_init_bootc.yaml
+++ b/.github/workflows/cloud_init_bootc.yaml
@@ -1,0 +1,77 @@
+name: Cloud Init Bootc image builds
+
+on:
+  schedule: # schedule the job to run at 12 AM daily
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM But assigned to Draft.

Gregory-Pereira:
I left it as a draft because in the process of working on this I discovered #431, which renders this pretty much useless as it will not work for both `amd-bootc` and `nvidia-bootc`

rhatdan:
@Gregory-Pereira Anything going on with this one?

Gregory-Pereira:
If there is a need I am happy to pick this up again. The E2E CI methodology no longer uses these images as we directly inject the SSH key. If you still think there is value in this feature beyond this I will pick it up again.

PR #432, opened 2024-05-02
fix(growfs): Correct the source path to be copied

YvanY0:
The current source path is `../../common/usr` and it will copy to destination path `build/usr/usr/`. I checked that the result are different between Linux and MacOS, so modify to adapt different platforms.

fix: #410

### DIFF
--- a/recipes/common/Makefile.common
+++ b/recipes/common/Makefile.common
@@ -159,7 +159,7 @@ install-chrome:
 .PHONY: growfs
 growfs: quadlet
 	# Add growfs service
-	mkdir -p build/usr; cp -R ../../common/usr/ build/usr/
+	mkdir build; cp -R ../../common/usr build/
 
 .PHONY: quadlet
 quadlet:


YvanY0:
Without this fix, the error is:
```
STEP 9/17: COPY build/usr/lib /usr/lib
Error: building at STEP "COPY build/usr/lib /usr/lib": checking on sources under "/home/yihyu/ai-lab-recipes/recipes/natural_language_processing/chatbot/bootc": copier: stat: "/build/usr/lib": no such file or directory

$ ls ai-lab-recipes/recipes/natural_language_processing/chatbot/build/usr/usr/lib
lib/     libexec/
```

lmilbaum (approved):
LGTM. Was this checked in CI?

rhatdan:
LGTM

YvanY0:
> LGTM. Was this checked in CI?

I tested this in a bare metal Linux system, and at that time I was very confused as to why it failed. So use the same `cp` command on my MacOS laptop, and the result was different, hence this PR.

It's worth adding Mac variant in the CI workflow, if someone can help to do that, that's awesome, or I can try to help with that when I have time. 😄

PR #433, opened 2024-05-02
Add EXTRA_RPM_PACKAGES to image Containerfile

javipolo:
With this change we can customize the image adding extra rpm packages

For example I've been using this in the past to install `cloud-init` into the new image so the user can customize the installed system via cloudinit in his cloud


lmilbaum (approved):
LGTM although it would be useful to provide the need that this contribution is addressing.

javipolo:
> LGTM although it would be useful to provide the need that this contribution is addressing.

I added an example of such use in the PR description, if you'd like that I also update the commit message I can do that

javipolo:
I updated the PR to include also the amd and intel images

training/amd-bootc/Containerfile
```diff
@@ -2,7 +2,10 @@ FROM quay.io/centos-bootc/centos-bootc:stream9
 
 ADD rocm.repo /etc/yum.repos.d/rocm.repo
 
-RUN dnf install -y rocm-smi && dnf clean all
+ARG EXTRA_RPM_PACKAGES=''
+RUN dnf install -y rocm-smi \
```
rhatdan: Doesn't
RUN dnf install -y rocm-smi ${EXTRA_RPM_PACKAGES}  && dnf clean all
Work?

training/intel-bootc/Containerfile
```diff
@@ -65,4 +66,6 @@ RUN if [ "${KERNEL_VERSION}" == "" ]; then \
     && mkdir -p /lib/modules/${KERNEL_VERSION}.${TARGET_ARCH}/extra/ \
     && mv /tmp/*.xz /lib/modules/${KERNEL_VERSION}.${TARGET_ARCH}/extra/ \
     && chown root:root /lib/modules/${KERNEL_VERSION}.${TARGET_ARCH}/extra/habanalabs*.xz \
-    && depmod -a ${KERNEL_VERSION}.${TARGET_ARCH}
+    && depmod -a ${KERNEL_VERSION}.${TARGET_ARCH} \
+    && xargs --no-run-if-empty dnf install -y <<< ${EXTRA_RPM_PACKAGES} \
```
rhatdan: ditto

training/nvidia-bootc/Containerfile
```diff
@@ -124,6 +126,7 @@ RUN if [ "${TARGET_ARCH}" == "" ]; then \
         dnf module enable -y nvidia-driver:${DRIVER_BRANCH} && \
         dnf install -y nvidia-fabric-manager-${DRIVER_VERSION} libnvidia-nscq-${DRIVER_BRANCH}-${DRIVER_VERSION} ; \
     fi \
+    && xargs --no-run-if-empty dnf install -y <<< ${EXTRA_RPM_PACKAGES} \
```
rhatdan: ditto

rhatdan:
I like the idea, but I think you can simplify the call.

training/amd-bootc/Containerfile
javipolo: it certainly will :) Gonna update it

training/intel-bootc/Containerfile
javipolo: In this case it wont work in case there are no EXTRA_RPM_PACKAGES, thus the `xargs --no-run-if-empty`

rhatdan:
```
$ cat /tmp/Containerfile 
from fedora
arg EXTRA_RPM_PACKAGES
run dnf -y install httpd ${EXTRA_RPM_PACKAGES}

$ podman build /tmp
WARN[0000] missing "EXTRA_RPM_PACKAGES" build argument. Try adding "--build-arg EXTRA_RPM_PACKAGES=<VALUE>" to the command line 
STEP 1/3: FROM fedora
...
Installed:
...                                        
  httpd-2.4.59-2.fc40.x86_64                                                    
  httpd-core-2.4.59-2.fc40.x86_64                                               
  httpd-filesystem-2.4.59-2.fc40.noarch                                         
  httpd-tools-2.4.59-2.fc40.x86_64                                              
...
Complete!
COMMIT
--> af67078efb7a
af67078efb7a2e1aeab860afdcc13b071a1c3804282318fd46044ce6ce38a2bd

 $ podman build --no-cache --build-arg EXTRA_RPM_PACKAGES=procps-ng /tmp | grep procps-ng
 procps-ng                      x86_64    4.0.4-3.fc40         fedora     367 k
(20/36): procps-ng-4.0.4-3.fc40.x86_64.rpm      3.6 MB/s | 367 kB     00:00    
  Installing       : procps-ng-4.0.4-3.fc40.x86_64                        36/36 
  Running scriptlet: procps-ng-4.0.4-3.fc40.x86_64                        36/36 
  procps-ng-4.0.4-3.fc40.x86_64  
```

This works for me?

javipolo:
> I like the idea, but I think you can simplify the call.

I could simplify it in two of the three Containerfiles. In the other, since there is no extra `dnf install` to piggyback onto, I think the xargs part is still needed

rhatdan:
arg EXTRA_RPM_PACKAGES=""
 Eliminates the warning.

rhatdan:
Much better and great idea.

LGTM

PR #437, opened 2024-05-02
Add growfs systemd service to bootc images

rhatdan:
This is a little hacky to just get this in.  Would be better if we shared this all with the recipes subdir, but for speed I just want to get this in.

We will need to revisit the Makefile.common concept to share more between recipes and training.


rhatdan:
@sallyom PTAL

sallyom (approved):
I'm ok with this for now - as soon as we have time we'll deduplicate things - also, I didn't build these images so if someone can confirm they build, merge it!

PR #439, opened 2024-05-02
intel training container is now pulling in vllm container

rhatdan:
In order to run podman inside of a container we need to disable SELinux enforcement and add CAP_SYS_ADMIN to allow mounting of overlay file systems. This matches what we are doing in the nvidia and amd bootc containers.


Gregory-Pereira (approved):
/lgtm

PR #456, opened 2024-05-03
Fix instructlab targets in README.md

markmc:


rhatdan:
Should probbaly allow both, fixed in the main branch and don't forget to sign your commits.

PR #467, opened 2024-05-04
Reboot instance after bootc install

lmilbaum:


Gregory-Pereira (approved):
/lgtm, any changes we can iterate on in the next one

PR #470, opened 2024-05-04
Allow both json and toml as config files for bootc-image-builder

javipolo:
With this change we can use both JSON and TOML files as bootc-image-builder configuration

Also I added a couple of variables to allow for easy extra customization, in case user wants to add some extra podman or bootc-image-builder arguments


rhatdan:
LGTM

Gregory-Pereira (approved):
/lgtm, hopefully we will catchup in CI soon and be able to test this.

PR #475, opened 2024-05-05
Add pull task to e2e workflow

Gregory-Pereira:
1) splits build into pull and build.
2) enables test escaping with hold tests.

### DIFF
--- a/training/tests/provision/playbook.yml
+++ b/training/tests/provision/playbook.yml
@@ -36,6 +36,7 @@
       authfile: /etc/containers/auth.json
 
   - name: Pull the Parent Image
+    async: 1000
     register: pull_result
     containers.podman.podman_image:
       name: "quay.io/ai-lab/{{ image_name }}:latest"
--- a/training/tests/provision/playbook.yml
+++ b/training/tests/provision/playbook.yml
@@ -35,6 +35,23 @@
       registry: quay.io
       authfile: /etc/containers/auth.json
 
+  - name: Pull the Parent Image
+    register: pull_result
+    containers.podman.podman_image:
+      name: "quay.io/ai-lab/{{ image_name }}:latest"
+      pull: true
+      auth_file: /etc/containers/auth.json
+      arch: amd64
+      state: present
+
+  - name: Check on Pulling the parent image
+    async_status:
+      jid: "{{ pull_result.ansible_job_id }}"
+    register: job_result
+    until: job_result.finished
+    retries: 100
+    delay: 10
+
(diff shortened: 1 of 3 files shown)


training/tests/provision/playbook.yml
```diff
@@ -35,6 +35,23 @@
       registry: quay.io
       authfile: /etc/containers/auth.json
 
+  - name: Pull the Parent Image
```
lmilbaum, line 38: async key is missing

lmilbaum (changes requested):
Should we block from pulling the image in the bootc install stage? Not sure if this was already implemented.

Gregory-Pereira:
Tests were held so I dont block your workstream on #471

Gregory-Pereira:
> Should we block from pulling the image in the bootc install stage? Not sure if this was already implemented.

Yes, let me add this.

training/tests/provision/playbook.yml
```diff
@@ -56,11 +80,14 @@
     poll: 0
     register: install_result
     ansible.builtin.shell: |
-      podman run --authfile=/etc/containers/auth.json \
-        --rm --privileged --pid=host \
```
Gregory-Pereira, line 60: @lmilbaum I saw this previously ran with `--rm` so I did not remove it, but it does not make sense to me why we want this. The instance will already clean itself up, this seems like just introducing another liability with respect to the `bootc paradigm`

training/tests/provision/playbook.yml
```diff
@@ -35,13 +35,38 @@
       registry: quay.io
       authfile: /etc/containers/auth.json
 
+  - name: Pull the Parent Image
```
lmilbaum, line 38: If the pull is flaky fail, what should happen?

training/tests/provision/playbook.yml
lmilbaum: Your argument makes sense but I'm not sure we want to touch it at this point

lmilbaum (changes requested):

training/tests/provision/playbook.yml
Gregory-Pereira: Then it will retry, or fail. In any case failing here is good to know, it helps us narrow down where this is breaking

Gregory-Pereira: Seems I misunderstood. Pull retry options coming in next set of tests

Gregory-Pereira:
Going to merge as this is stable.

PR #476, opened 2024-05-05
ansible cfg refactor for e2e training workflow

Gregory-Pereira:
pulls ansible cfg refactor out of #475 
/cc @lmilbaum


lmilbaum (approved):
I like this refactor

PR #494, opened 2024-05-15
add interrupt to llamacpp

MichaelClifford:
Partially addresses https://github.com/containers/ai-lab-recipes/issues/497

This will prevent the llamacpp_python model server from crashing if you send a second request while the model is streaming output. It seems to still crash if a second request is sent while the model is processing the first.


Gregory-Pereira (approved):
/lgtm

PR #495, opened 2024-05-15
Lower ConversationBufferWindowMemory

MichaelClifford:
the ConversationBufferWindowMemory size was set too high, letting the conversation history reach the context limit before clearing out earlier messages, and causing the server to crash.  This PR lowers the value so that we only retain the last 3 chat messages in the history. 

Should address  https://github.com/containers/podman-desktop-extension-ai-lab/issues/1083


Gregory-Pereira (approved):
/lgtm

PR #507, opened 2024-05-22
debugging e2e training workflow

Gregory-Pereira:
Now that https://github.com/instructlab/instructlab/pull/1185 has landed we are unblocked in testing e2e workflow.


lmilbaum:
CI is broken

PR #509, opened 2024-05-23
Fix Ilab container builds

Gregory-Pereira:
Reverting to least blocked action run: https://github.com/containers-mirror/ai-lab-recipes/actions/runs/8903515843/job/24451391934#step:4:32

### DIFF
--- a/.github/workflows/instructlab.yaml
+++ b/.github/workflows/instructlab.yaml
@@ -4,18 +4,18 @@ on:
   schedule: # schedule the job to run at 12 AM daily
    - cron: '0 0 * * *'
 
-  # pull_request:
-  #   branches:
-  #     - main
-  #   paths:
-  #     - .github/workflows/instructlab_baseimages_build_push.yaml
-  #     - training/nvidia
+  pull_request:
+    branches:
+      - main
+    paths:
+      - .github/workflows/instructlab.yaml
+      - training/instructlab/Makefile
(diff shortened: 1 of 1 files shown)


markmc:
Are you intentionally reverting @rhatdan's move to OCI transport in #427 ?

Gregory-Pereira:
Yes I was. Because that change broke all of our workflows which automatically produce our ai-lab ilab based images. Because of this our e2e runs have not worked. Although I guess a potential alternative implementation would be swapping CI to work with an action build rather than a `make` command ... But I wanted to keep our automation + CI as closely aligned to our average user experience as possible.

rhatdan:
We need these to be stored as OCI images for speed of building and saving of disk space. Lets work on some other mechanism to fix the ci sysem.

Gregory-Pereira:
Understood, ill update my implementation shortly.

lmilbaum:
CI is broken

cooktheryan:
@rhatdan @lmilbaum @Gregory-Pereira in regards to the e2e/ilab testing will we keep maintaining this or will this go into official repos such as the rhelai repository?

rhatdan:
I think we should keep the project as an open project, so the Training bits should stay here.

cooktheryan:
Will try to prioritize the stability of the e2e tests then next week

PR #510, opened 2024-05-23
Fix invalid instructions in README.md

gcolman:
Fixed two instructions in the README that stop the examples working. 
1) the instruction to make model pointed to etr-resnet-50 rather than the etr-resnet-101 that the instructions use.  2) The client container start had a /detecion in the model address where it should not have.


Gregory-Pereira (approved):
The `README.md`s sometimes get out of date because we usually maintain the Makefiles as the source of truth. Both changes look good, thanks!

Gregory-Pereira:
Please make sure to just signoff your commit to pass DCO.

rhatdan:
LGTM
git commit -a --amend -s
git push --force

And then we can merge.

PR #511, opened 2024-05-23
Simplify package installation

enriquebelarte:
Installs epel and CRB in a cleaner way.


lmilbaum (approved):
LGTM

lmilbaum:
Is it an option to use variables instead of if/else commands?

enriquebelarte:
> Is it an option to use variables instead of if/else commands?

Maybe for the `KERNEL_VERSION`, `TARGET_ARCH` and `OS_VERSION_MAJOR` we could leave out the conditionals (IINM there's no `else`) and add variables which could be cleaner and better for CI ? .  Regarding the packages based on different OSes I think it's OK as it is now, but I accept suggestions.

PR #516, opened 2024-05-29
bump python to 3.12 version in rag workflow

Gregory-Pereira:
/cc @lmilbaum @MichaelClifford


lmilbaum (changes requested):
Bumping the python version could be helpful (how come renovate didn't bump it). But, the order change won't impact the build image step. Would suggest to keep the step where it was and maybe bump the python version.

Gregory-Pereira:
Appreciate the clarification, I misunderstood what I was bumping. This has been moved to #515 commit 2. Closing.

PR #518, opened 2024-05-30
Add initial tool for creating custom eval sets

MichaelClifford:
Wrote a small streamlit app that can help users create custom evaluation sets for promptfoo and their specific LLM use cases with some helpful visualization. I'm sure it can be improved moving forward, but this seems like a good first step. 

The tool allows you to define a question, then define 3 correct and 3 incorrect answers. It then plots their cosine similarities from the question. This can be used as a proxy for correctness. Once the users feels good about their evaluation set (All right answers clustered near the question and all wrong answers clustered away from the question). They can submit the question to an LLM, and plot the answer on the same graph to see how close it is to the question and other right answers. We can use where the LLM answer falls along the spectrum of right and wrong answers as a proxy for correctness.    
  
Screenshot 2024-05-30 at 4 43 07 PM

This tool should make crating custom evaluation sets for models easier.

### DIFF
--- a/eval/embeddings/custom_eval_set.py
+++ b/eval/embeddings/custom_eval_set.py
@@ -1,0 +1,98 @@
+from langchain.embeddings.sentence_transformer import SentenceTransformerEmbeddings
+from langchain_openai import ChatOpenAI
+from langchain.chains import LLMChain
+from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder
+import matplotlib.pyplot as plt
+import os
+from scipy.spatial.distance import cosine
+import streamlit as st
+
+
+model_service = os.getenv("MODEL_ENDPOINT",
+                          "http://localhost:8001")
+model_service = f"{model_service}/v1"
+
+embedding_model = os.getenv("EMBEDDING_MODEL",
+                            "BAAI/bge-base-en-v1.5")
+
+def get_embedding(string, e):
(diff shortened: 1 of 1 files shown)


cooktheryan:
@Shreyanand do you want to take a look at this?

sallyom (commented):
LGTM but needs a document that has the contents of the PR description, at least - this can be a follow-up since Michael is off for a bit.

eval/embeddings/custom_eval_set.py
```diff
@@ -0,0 +1,98 @@
...
+
+    chain = LLMChain(llm=llm, 
+                prompt=prompt,
+                verbose=False,)
+     
+    response = chain.invoke(st.session_state["Question"])
+    st.session_state["Answers"]["LLM Response"] = response["text"]
+    st.markdown(st.session_state["Answers"]["LLM Response"])
+    st.rerun()
+
```
lmilbaum, line 98: Please fix end of file

lmilbaum:
A general comment. Does this tool require tests to be invoked in CI? Maybe unit test would suffice as a first step. WDYT?

Shreyanand (commented):
LGTM 👍 
I'll test, improve, and document it in a follow up PR.

rhatdan:
Since I believe @MichaelClifford is away, I will merge and can @Shreyanand cleanup @lmilbaum issue.

PR #519, opened 2024-05-31
Configure model for chatbot and codegen recipes

johnmcollier:
A couple small changes to the codegen and chatbot recipes, after doing some work with them using a vLLM model service:
- Pass in model name to the llm object in `codegen-app.py`
- Allow environment variable to configure model name in both `chatbot_ui.py` and `codegen-app.py`

### DIFF
--- a/recipes/natural_language_processing/chatbot/app/chatbot_ui.py
+++ b/recipes/natural_language_processing/chatbot/app/chatbot_ui.py
@@ -67,7 +67,7 @@ def memory():
     memory = ConversationBufferWindowMemory(return_messages=True,k=3)
     return memory
 
-model_name = "" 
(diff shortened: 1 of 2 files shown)


recipes/natural_language_processing/chatbot/app/chatbot_ui.py
```diff
@@ -67,7 +67,7 @@ def memory():
     memory = ConversationBufferWindowMemory(return_messages=True,k=3)
     return memory
 
-model_name = ""
```
johnmcollier, line 70: A better approach might be to add in a selector for the model like was done for Ollama, but I wanted to favour a smaller and simpler change to start with

sallyom (approved):
LGTM

rhatdan:
LGTM

PR #523, opened 2024-05-31
adding single concurrency value to e2e

Gregory-Pereira:
Redo #522 against upstream for secret access


Gregory-Pereira:
Solves VPC collisioncs, cc @lmilbaum

PR #529, opened 2024-06-03
fix of arm err on previous run

cooktheryan:
On previous run https://github.com/containers/ai-lab-recipes/actions/runs/9352483549/job/25740755514 we had a failure due to a missing package. This will attempt to resolve that issue

### DIFF
--- a/recipes/natural_language_processing/chatbot/provision/playbook.yml
+++ b/recipes/natural_language_processing/chatbot/provision/playbook.yml
@@ -6,19 +6,22 @@
   gather_facts: false
 
   tasks:
-  
+
   - name: Wait until the instance is ready
     ansible.builtin.wait_for_connection:
     delay: 10
     timeout: 60
-  
+
   - name: Gather facts for first time
     ansible.builtin.setup:
 
   - name: Required Packages
     ansible.builtin.package:
-      name: podman
+      name: "{{ item }}"
       state: present
+    with_items:
+    - podman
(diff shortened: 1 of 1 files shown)


lmilbaum (changes requested):
Could you please provide additional context for this change? Which pipeline was failing due to the missing package?

cooktheryan:
@lmilbaum sorry about that. I wonder if it would be worth it for this PR to try to test the job as well. I could make a modification to that job to run if the files change

.github/workflows/testing_framework.yaml
```diff
@@ -1,9 +1,13 @@
 name: Testing Framework
 
 on:
+  push:
```
lmilbaum: That won't work because the containers org doesn't have the right runner for running the workflows. It is running on the containers-mirror/ai-lab-recipes repo after the pushed commit is synced.

lmilbaum (changes requested):

.github/workflows/testing_framework.yaml
cooktheryan: k removing

rhatdan:
Can you squash your commits?

cooktheryan:
@rhatdan squashed

rhatdan (approved):

lmilbaum (approved):

Gregory-Pereira (approved):

PR #533, opened 2024-06-06
need to pin version specific version of fedora

cooktheryan:
Pin version due to uncertainty on Fedora version that may be running in tests


Gregory-Pereira (approved):
/lgtm
One question -- does this need to be implemented on the other Github workflows? Or should we open separate PRs for those because AMIs would be different?

cooktheryan:
Separate where applicable. One of the other tests uses RHEL and 9.4 specifically defined. Will open additional prs if I find additional scaries

PR #534, opened 2024-06-06
Improve Makefiles

rhatdan:


rhatdan:
@cgwalters PTAL

rhatdan:
@lmilbaum @sallyom @Gregory-Pereira @cooktheryan PTAL

Gregory-Pereira (approved):
/lgtm, let me know what of these training images we want to see worked into CI

rhatdan:
The training objects take a long time to process and move a ton of data around.  So you could try with deepspeed and vllm to start.

PR #528, opened 2024-06-03
lamacpp-python-vulkan for arm64

Gregory-Pereira:
Addresses #526 cc @jeffmaury 
Let me know if I misunderstood anything in my assessment of your issue.

PR #535, opened 2024-06-06
fix of broken install

cooktheryan:
This should fix the errors we are seeing with this test 

https://github.com/containers/ai-lab-recipes/pull/528


rhatdan (approved):
LGTM

Issue #526, opened 2024-06-03
https://quay.io/repository/ai-lab/llamacpp-python-vulkan is provided only for aarch64


jeffmaury:

Gregory-Pereira:
~~I think you mean its only provided for `amd64` and not for `arm64`. I will add a build for it now.~~ I was looking at the Cuda image workflow, however this is the vulkan image and it was pushed by hand instead of workflow, updating the CI for this.

Gregory-Pereira:
So it turns out that the container file is hardcoded for installation on AARCH64, it needs to be re-written to support amd64. That makes this a 2 step process, the original PR I will rework to just fix automated builds on Aarch64 and then a second one to refactor the containerfile to build for both.

Gregory-Pereira:
Code has been merged, images coming soon.

PR #540, opened 2024-06-08
vulkan amd64 build

Gregory-Pereira:
Build works for me.

Considering some other copr repos for the `vulkan-tools` and `vulkan-headers`:

- `ligenix/enterprise-sandbox`
- `kasion/Hyprland-nightly`
- `mskarbek/swaywm`

Addresses: #526


Gregory-Pereira:
moved to #541 so that tests can run with secrets on merge.

PR #542, opened 2024-06-09
Fix Makefiles to allow alternative container engines

rhatdan:
Some users want to use buildah instead of podman to build their container engines.

Buildah does not support --squash-all but after examining the podman code --squash-all ends up just being the equivalent of "--squash --layers=false"

### DIFF
--- a/recipes/natural_language_processing/rag/Makefile
+++ b/recipes/natural_language_processing/rag/Makefile
@@ -16,7 +16,7 @@ run:
 # rag requires custom bootc because it uses an extra build-arg for CHROMADB_IMAGE (other apps use ../../common/Makefile.common target)
 .PHONY: bootc
 bootc: quadlet
-	podman build \
+	"${CONTAINER_TOOL}"  build \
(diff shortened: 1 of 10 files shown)


recipes/natural_language_processing/rag/Makefile
```diff
@@ -16,7 +16,7 @@ run:
 # rag requires custom bootc because it uses an extra build-arg for CHROMADB_IMAGE (other apps use ../../common/Makefile.common target)
 .PHONY: bootc
 bootc: quadlet
-	podman build \
+	"${CONTAINER_TOOL}"  build \
```
Gregory-Pereira, line 19: Did we want to include the `--layers=false` in this bootc command?

Gregory-Pereira (approved):
/lgtm just one question

recipes/natural_language_processing/rag/Makefile
rhatdan: We are not squashing so no.

PR #545, opened 2024-06-10
podman build does not support --squash and --layers at same time

rhatdan:
Fixed: https://github.com/containers/podman/pull/22957


lmilbaum (changes requested):
None of the training workflows was triggered in CI. Please unlock the CI such that this PR can be verified before it is merged.

PR #549, opened 2024-06-11
trying to pass runtime var to allow for model pull

cooktheryan:
We need to login to HF to be able to pull Mixtral https://github.com/containers/ai-lab-recipes/actions/runs/9465215740

This should hopefully fix it. If not the follow up PR(s) will


Gregory-Pereira (approved):
/lgtm

PR #557, opened 2024-06-13
Add tmux to the bootc image

fabiendupont:
Many commands that are run for SDG and training can take a lot of time, so there is a risk to have a network disconnection during the task. With `tmux`, users have the ability to detach the jobs from their SSH session and let the tasks run for a very long time.

### DIFF
--- a/training/intel-bootc/Containerfile
+++ b/training/intel-bootc/Containerfile
@@ -24,10 +24,10 @@ RUN if [ -f /etc/centos-release ]; then \
        dnf -y update \
        && dnf -y install epel-release \
        && crb enable \
-       && dnf -y install ninja-build pandoc pciutils;\
+       && dnf -y install ninja-build pandoc pciutils tmux;\
     fi \
     && dnf -y update \
-    && dnf install -y dkms ninja-build pandoc pciutils \
+    && dnf install -y dkms ninja-build pandoc pciutils tmux \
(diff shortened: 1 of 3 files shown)


training/intel-bootc/Containerfile
```diff
@@ -24,10 +24,10 @@ RUN if [ -f /etc/centos-release ]; then \
        dnf -y update \
        && dnf -y install epel-release \
        && crb enable \
-       && dnf -y install ninja-build pandoc pciutils;\
+       && dnf -y install ninja-build pandoc pciutils tmux;\
     fi \
     && dnf -y update \
-    && dnf install -y dkms ninja-build pandoc pciutils \
+    && dnf install -y dkms ninja-build pandoc pciutils tmux \
```
lmilbaum, line 30: A non blocking comment. This command is executed twice if it is a centos based image.

lmilbaum (approved):
LGTM

lmilbaum:
The CI didn't trigger the relevant pipelines to validate the changes. I'm aware that it shouldn't be in the scope of this PR but it would be very valuable if the CI provides valuable input before this PR is merged.

training/intel-bootc/Containerfile
cooktheryan: +1 we should just add an else but can come back to it

PR #558, opened 2024-06-13
Fix typo for pciutils in AMD bootc

fabiendupont:


lmilbaum (approved):
JFYI, CI didn't trigger valuable pipeline to validate this change.

Gregory-Pereira:
Are you pointing out that this is unexpected behaviour @lmilbaum, or just informing him that this wont run the testing suite? I was under the assumption we wanted to run these heavy duty pipelines only on a crontab per day.

PR #563, opened 2024-06-14
Add BASEIMAGE variable

enriquebelarte:
Adding BASEIMAGE to Containerfile.builder allows using different images for building


rhatdan:
This is fine, but we need to do the same for intel and amd.

PR #572, opened 2024-06-17
Make growfs conditional

enriquebelarte:
growfs is created by Makefile and CI does not use it. Also if I'm not misktaken growfs is only used for disk images creation. By changing this, growfs files will only be created when `Makefile` is run, so CI build pipelines can use the `Containerfile` and growfs can be also used when needed running `make`.

If change is accepted I'll submit additional PRs for amd and intel.


cooktheryan:
/LGTM

@Gregory-Pereira thoughts?

Gregory-Pereira (approved):
/lgtm

rhatdan:
This is wrong.

PR #575, opened 2024-06-18
Make growfs conditional

enriquebelarte:
As explained at https://github.com/containers/ai-lab-recipes/pull/572 copy for growfs function just if run from `Makefile`

### DIFF
--- a/training/amd-bootc/Containerfile
+++ b/training/amd-bootc/Containerfile
@@ -3,8 +3,8 @@ FROM ${BASEIMAGE}
 
 ADD rocm.repo /etc/yum.repos.d/rocm.repo
 
-# Include growfs service
-COPY build/usr /usr
+# Include growfs service if needed
+RUN if [ -d "build/usr" ]; then cp -r build/usr /usr; fi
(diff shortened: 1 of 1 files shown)


training/amd-bootc/Containerfile
```diff
@@ -3,8 +3,8 @@ FROM ${BASEIMAGE}
 
 ADD rocm.repo /etc/yum.repos.d/rocm.repo
 
-# Include growfs service
-COPY build/usr /usr
```
rhatdan, line 7: These are not the same thing.

COPY build/usr /usr

Will copy outside of the container from the host into /usr in the container.

cp -r build/usr /usr 

Will copy content from /build/usr to /usr from within the container image.

enriquebelarte: You are completely right Dan. Have to find out a way to make it work from Konflux and undo other PRs with the same change.
Will close it.

rhatdan:
Replaced with https://github.com/containers/ai-lab-recipes/pull/589

PR #585, opened 2024-06-19
Use multi-stage to build the Intel bootc image

fabiendupont:
The NVIDIA bootc container is using multi-stage to avoid shipping build dependencies in the final image, making it also smaller. This change implements the same build strategy for the Intel bootc image.

The builder image is the same as for NVIDIA bootc. It is currently named after NVIDIA, but should be renamed in a follow-up change. The benefit is that a single builder image is maintained for all bootc images that require out-of-tree drivers.

The number of build arguments is also reduced, since most of the information is already present in the builder image. There is only one kernel package per builder image and one image per architecture, so we can retrieve the `KERNEL_VERSION` and `TARGET_ARCH` variables by querying the RPM database. The OS information is retrieved by sourcing the `/etc/os-release` file.

The extraction of the RPMs doesn't require storing the files, as `rpm2cpio` supports streaming the file over HTTP(S). This number of commands is smaller and the downloads happened already for each build, since the download was not in a separate `RUN` statement.

It is not necessary to copy the source of the drivers in `/usr/src\, since we don't need to keep it in the final image. The Makefiles accept a `KVERSION` variable to specify the version of the kernel and resolve its path. The other benefit is to build as non-root.

The `.ko` files can then be copied to the final image with `COPY --from=builder`. The change also ensures that the firmware files are copied to the final image.

This change also adds support for `EXTRA_RPM_PACKAGES`.

### DIFF
--- a/training/intel-bootc/Containerfile
+++ b/training/intel-bootc/Containerfile
@@ -1,61 +1,55 @@
+ARG DRIVER_TOOLKIT_IMAGE="quay.io/ai-lab/nvidia-builder:latest"
 ARG BASEIMAGE="quay.io/centos-bootc/centos-bootc:stream9"
-FROM ${BASEIMAGE}
 
-ARG OS_VERSION_MAJOR=''
-ARG DRIVER_VERSION=1.16.0-526
-ARG TARGET_ARCH=''
-ARG KERNEL_VERSION=''
-ARG REDHAT_VERSION='el9'
+FROM ${DRIVER_TOOLKIT_IMAGE} as builder
+
+ARG DRIVER_VERSION=1.16.1-7
 ARG HABANA_REPO="https://vault.habana.ai/artifactory/rhel/9/9.2"
-ARG KERNEL_DIR="/usr/lib/modules/${KERNEL_VERSION}.el9_4.x86_64/build"
-ARG KERNEL_FULL_VER="${KERNEL_VERSION}.el9_4.x86"
 
+WORKDIR /home/builder
 
 RUN . /etc/os-release \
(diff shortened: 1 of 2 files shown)


training/intel-bootc/Containerfile
```diff
@@ -1,62 +1,58 @@
+ARG DRIVER_TOOLKIT_IMAGE="quay.io/ai-lab/nvidia-builder:latest"
```
enriquebelarte, line 1:
```suggestion
ARG DRIVER_TOOLKIT_IMAGE="quay.io/ai-lab/intel-builder:latest"
```

Gregory-Pereira: I was under the assumption this had to do with what he explained in the comment, where we will be starting with the `nvidia-builder` image name, and then would do a follow on PR to rename it to just `builder` as "it is currently named after NVIDIA, but should be renamed in a follow-up change. The benefit is that a single builder image is maintained for all bootc images that require out-of-tree drivers."

enriquebelarte: Even if I prefer a builder for each driver I would change the name of the builder to something like `common-builder` to avoid confusion.

fabiendupont: Does that image exist today? I mention that it should become a unique image not related to a vendor, but in a follow PR.

Gregory-Pereira: To clarify this nothing currently exists in quay, we've only been using local builds of dtk

Gregory-Pereira (approved):
/lgtm

PR #589, opened 2024-06-20
Eliminate growfs from training containers

rhatdan:
Training images will not be installed in the cloud by default so should not include growfs.

### DIFF
--- a/training/amd-bootc/Containerfile
+++ b/training/amd-bootc/Containerfile
@@ -10,9 +10,6 @@ FROM ${BASEIMAGE}
 
 ADD rocm.repo /etc/yum.repos.d/rocm.repo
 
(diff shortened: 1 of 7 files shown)


rhatdan:
@enriquebelarte  PTAL

enriquebelarte:
LGTM. Thank you.

Gregory-Pereira:
Should this then be baked into the `cloudinit` images? Still not sure if those are a thing (I just see the directory still in tree) so if they are I can make a follow on issue for this.

rhatdan:
Currently not supported, but yes we should add those to cloudinit images, although Cloudinit images might automatically do the growfs part. @cgwalters would know for sure.

PR #590, opened 2024-06-20
Fix image builder

javipolo:
- Properly separate and order podman and bootc-image-builder arguments
- Move all the `selinux.tmp` workaround to the same layer, so bootc install wont complain about missing files


rhatdan (approved):
LGTM

PR #591, opened 2024-06-20
No longer using vllm and deepspeed images for AI bootc images

rhatdan:
Instructlab will include these executables inside of its container rather then using external containers.

### DIFF
--- a/training/Makefile
+++ b/training/Makefile
@@ -3,7 +3,7 @@ MAKEFLAGS += -j2
 default: help
 
 .PHONY: all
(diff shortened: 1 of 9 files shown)


rhatdan:
@n1hility PTAL

n1hility (approved):

n1hility:
LGTM . Well also need to update the wrapper script to just run everything through the ilab container but that can be a follow up i can do that quick after

rhatdan:
SGTM

Issue #592, opened 2024-06-20
'make amd' fails with "Error: short-name resolution enforced but cannot prompt without a TTY"


prarit:
I am running on RHEL9.4.  The amd target fails as follows:

```
[01:17 PM root@intel-eaglestream-spr-17 training]# make amd
make -C amd-bootc/ bootc bootc-models
make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
make[1]: Entering directory '/root/ai-lab-recipes/training/amd-bootc'
mkdir -p /root/ai-lab-recipes/training/amd-bootc/../build
cp -pf /root/ai-lab-recipes/training/amd-bootc/../ilab-wrapper/ilab /root/ai-lab-recipes/training/amd-bootc/../build
cp -pf /root/ai-lab-recipes/training/amd-bootc/../ilab-wrapper/ilab-qlora /root/ai-lab-recipes/training/amd-bootc/../build
cp -pf /root/ai-lab-recipes/training/amd-bootc/../ilab-wrapper/ilab-training-launcher /root/ai-lab-recipes/training/amd-bootc/../build
# Add growfs service
mkdir -p build; cp -pR ../common/usr build
"podman" build \
	 \
	 \
	 \
	 \
	--build-arg INSTRUCTLAB_IMAGE=quay.io/ai-lab/instructlab-amd:latest \
	--timestamp=1718896777 \
	--build-arg VLLM_IMAGE=quay.io/ai-lab/vllm:latest \
[...]

PR #593, opened 2024-06-20
make sure downloader image is present and called non short name

Gregory-Pereira:
Addresses: https://github.com/containers/ai-lab-recipes/issues/592
Please review @n1hility @rhatdan 
cc @cooktheryan @prarit

Change: model downloader is not present as well as short name resolution (not sure if this is a problem yet until it becomes available)

### DIFF
--- a/training/Makefile
+++ b/training/Makefile
@@ -67,7 +67,7 @@ vllm:
 driver-tookit:
 	make -C common/ -f Makefile.common driver-toolkit
 amd:
-	make -C amd-bootc/ bootc bootc-models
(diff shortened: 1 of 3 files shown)


rhatdan:
I am working on a similar fix for this.  Want to add `--pull never` to the podman run commands.

rhatdan:
I was putting the common code into ../common/

Gregory-Pereira:
Ive both tagged this as `local` instead of `latest` as well as add the pull never

rhatdan:
I now like mine better, lets continue on that one.

PR #594, opened 2024-06-20
Fix up pulling images

rhatdan:


rhatdan:
Alternative to https://github.com/containers/ai-lab-recipes/pull/593

@n1hility PTAL

rhatdan:
@Gregory-Pereira PTAL

Gregory-Pereira:
Changes look good to me. Prarit is testing from #593 which seems to be a dupe of this. Happy to use this instead but going to leave #593 up for him to continue testing there then we can merge this. Regardless approving for now.

Gregory-Pereira (approved):
see above comment

rhatdan:
I think we should move more stuff to the model package, and maybe go with yours.

rhatdan:
@Gregory-Pereira Ok I reworked it again to put all of the model commands in the model subdir.  I think this is the way we should go.

rhatdan:
This is blowing up on me because of permissions.

```Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/huggingface_hub/utils/_errors.py", line 304, in hf_raise_for_status
    response.raise_for_status()
  File "/usr/lib/python3.9/site-packages/requests/models.py", line 943, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://huggingface.co/api/models/mistralai/Mixtral-8x7B-Instruct-v0.1/revision/main

The above exception was the direct cause of the following exception:
```

training/model/Makefile
```diff
@@ -20,4 +20,29 @@ download: image
 .PHONY: download-model
 download-model:
 	mkdir -p ../build/models
-	podman run -e HF_TOKEN -v ../build/models:/download:z -t model-downloader huggingface-cli download --exclude "*.pt" --local-dir "/download/$(MODEL_REPO)" "$(MODEL_REPO)"
+	podman run \
+		-e HF_TOKEN \
```
sallyom, line 24: where's the HF_TOKEN set?

sallyom: nm, i see it in the .github/workflow

Gregory-Pereira:
Its not permissions its an outage

training/model/Makefile
Gregory-Pereira: but dont we need to call it as an env variable tho like: `-e HF_TOKEN=$$HF_TOKEN` or as a make value passed: `-e HF_TOKEN=$(HF_TOKEN)`?

PR #596, opened 2024-06-20
Top level bootc with images

Gregory-Pereira:
/cc @n1hility @rhatdan 

Changes:

- `HF_TOKEN` gets taken from env (before I think its passing blank value)
- Top level commands for bootc + models

### DIFF
--- a/training/model/Containerfile
+++ b/training/model/Containerfile
@@ -1,3 +1,4 @@
 FROM registry.access.redhat.com/ubi9/ubi
+RUN if [ -z $HF_TOKEN ]; then; echo "Error. Please set your \$HF_TOKEN in env. Required to pull mixtral."; exit 1; fi;
 RUN dnf install -y python3-pip && python3 -m pip install huggingface_hub[cli]
 WORKDIR /download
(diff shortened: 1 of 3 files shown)


training/model/Containerfile
```diff
@@ -1,3 +1,4 @@
 FROM registry.access.redhat.com/ubi9/ubi
+RUN if [ -z $HF_TOKEN ]; then; exit 1; fi;
```
rhatdan: You should print an error message to the user telling them that this is required.

training/Makefile
```diff
@@ -75,6 +75,18 @@ intel: bootc-models
 nvidia: bootc-models
 	make -C nvidia-bootc/ driver-toolkit bootc
 
+#
+# Create bootc container images prepared with AI models pre-loaded
+#
+
+.PHONY: amd-model nvidia-model intel-model 
+amd-model:
+	make -C bootc-models amd
```
rhatdan: I think this is wrong there is no bootc-models subdir?

n1hility (approved):

PR #599, opened 2024-06-21
Change source from ilab wrapper

enriquebelarte:
Copying the `ilab` script from the image mount as it is done now breaks Konflux CI. If I'm not missing something we could get the same existing script from upper level and also that would ease CI tasks.
/cc @rhatdan @fabiendupont @Gregory-Pereira

### DIFF
--- a/training/amd-bootc/Containerfile
+++ b/training/amd-bootc/Containerfile
@@ -22,9 +22,10 @@ RUN dnf install -y \
 # Setup /usr/lib/containers/storage as an additional store for images.
 # Remove once the base images have this set by default.
 RUN sed -i -e '/additionalimage.*/a "/usr/lib/containers/storage",' \
-	/etc/containers/storage.conf && \
-	cp /run/.input/ilab /usr/local/bin/ilab
+	/etc/containers/storage.conf 
 
+COPY ../ilab-wrapper/ilab /usr/local/bin/ilab
+ 	
 ARG INSTRUCTLAB_IMAGE="quay.io/ai-lab/instructlab-amd:latest"
 
 ARG SSHPUBKEY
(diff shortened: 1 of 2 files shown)


n1hility (approved):
LGTM

training/amd-bootc/Containerfile
```diff
@@ -22,9 +22,10 @@ RUN dnf install -y \
 # Setup /usr/lib/containers/storage as an additional store for images.
 # Remove once the base images have this set by default.
 RUN sed -i -e '/additionalimage.*/a "/usr/lib/containers/storage",' \
-	/etc/containers/storage.conf && \
-	cp /run/.input/ilab /usr/local/bin/ilab
+	/etc/containers/storage.conf 
 
+COPY ../ilab-wrapper/ilab /usr/local/bin/ilab
```
Gregory-Pereira, line 27: Please correct me if I am wrong, but shouldn't we pull in `instructlab-amd` as `ilab` for the AMD bootc image -->
`COPY ../ilab-wrapper/insteuctlab-amd /usr/local/bin/ilab` or are they separate binaries?

training/nvidia-bootc/Containerfile
```diff
@@ -156,8 +156,9 @@ fi
 # Also make sure not to duplicate if a base image already has it specified.
 RUN grep -q /usr/lib/containers/storage /etc/containers/storage.conf || \
     sed -i -e '/additionalimage.*/a "/usr/lib/containers/storage",' \
-	/etc/containers/storage.conf && \
-	cp /run/.input/ilab /usr/local/bin/ilab
+	/etc/containers/storage.conf 
+
+COPY ../ilab-wrapper/ilab /usr/local/bin/ilab
```
Gregory-Pereira, line 161: Same as above.

Gregory-Pereira (changes requested):

training/amd-bootc/Containerfile
n1hility: We don't (yet) have an amd version, so that file doesn't exist yet, but in concept yes you are correct.

Gregory-Pereira (approved):

n1hility:
Ah darn we are going to need to revert this:

Error: building at STEP "COPY ../ilab-wrapper/ilab /usr/local/bin/ilab": checking on sources under "/mnt/disk/ai-lab-recipes/training/nvidia-bootc": possible escaping context directory error: copier: stat: "/ilab-wrapper/ilab": no such file or directory

PR #603, opened 2024-06-21
Bring your own podman 5

cooktheryan:
@Gregory-Pereira we did this for Fetch it to allow for us to force update the podman version. This may allow for us to get around the missing functionality


cooktheryan:
Got the approval from greg in chat. Going to test on the other side

PR #614, opened 2024-06-24
moving tests from mirror to regular

Gregory-Pereira:
Now we have updated our test infrastructure, we should run tests against our repo rather than the mirror. The benefits of this are that now people dont have to have code merged to test it.
cc @lmilbaum @cooktheryan


rhatdan (approved):
LGTM

PR #628, opened 2024-06-25
Use --installed to get kernel version for packages to be installed

javipolo:
This fix should pin the kernel version to the one of the installed kernel-core package, which should always be the same one

This will solve when sometimes the modules are built for different versions, or other weird things:

```
[root@localhost modules]# find . | grep nvidia
./5.14.0-427.13.1.el9_4.x86_64/kernel/drivers/platform/x86/nvidia-wmi-ec-backlight.ko.xz
./5.14.0-427.22.1.el9_4.x86_64/extra/drivers/video/nvidia
./5.14.0-427.22.1.el9_4.x86_64/extra/drivers/video/nvidia/nvidia-drm.ko
```

### DIFF
--- a/training/common/driver-toolkit/Containerfile
+++ b/training/common/driver-toolkit/Containerfile
@@ -8,9 +8,8 @@ ARG ENABLE_RT=''
 USER root
 
 RUN if [ "${KERNEL_VERSION}" == "" ]; then \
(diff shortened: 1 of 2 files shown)


rhatdan (approved):
LGTM

PR #559, opened 2024-06-13
Temporary fix for bootc CI builds

enriquebelarte:
Konflux CI fails when building using bootc images as base throwing this error:
`Error: Cannot create repo temporary directory "/var/cache/dnf/baseos-044cae74d71fe9ea/libdnf.1jsyRp": Permission denied`
 This temporary workaround is needed for build pipeline to work on Konflux CI until libdnf fix is merged to RHEL. 
References:
https://issues.redhat.com/browse/RHEL-39796
https://github.com/rpm-software-management/libdnf/pull/1665

It should be removed once the permanent fix is merged.

PR #630, opened 2024-06-25
Temp workaround for amd-bootc

enriquebelarte:
Same as in https://github.com/containers/ai-lab-recipes/pull/559 we need this temporary workaround to bypass `libdnf `issue when installing packages.


Gregory-Pereira (approved):
/lgtm

PR #631, opened 2024-06-25
Fix model targets

javipolo:
A couple of fixes so Make targets in models/ do not break

The PR https://github.com/containers/ai-lab-recipes/pull/611 also has those changes, but it includes many more things and it is currently on hold

If the other PR is going to stay on hold, this should be merged ASAP to fix the currently broken pipeline


rhatdan (approved):
LGTM

PR #635, opened 2024-06-26
Add check for installed kernel-core

enriquebelarte:
Latest changes broke with some images that lack kernel-core package installed by default: i.e. ubi9.
Thanks @javipolo for your suggestion.


cooktheryan:
you can find the error shown here https://github.com/containers/ai-lab-recipes/actions/runs/9681290098/job/26711665880

PR #636, opened 2024-06-26
without the kernel-core package build fails

cooktheryan:
Currently testing is broken because the centos stream does not have the package that allows for the kernel version to be defined

https://github.com/containers/ai-lab-recipes/actions/runs/9681290098/job/26711665880


cooktheryan:
@enriquebelarte 's pull request is better

PR #637, opened 2024-06-26
Exclude kernel packages from dnf update

javipolo:
dnf update was there to update python3 packages because of CVE: https://catalog.redhat.com/software/containers/ubi9/ubi/615bcf606feffc5384e8452e?architecture=amd64&image=66610af1e7e2103fc98023b8&container-tabs=security

Also `dnf update` might install a newer kernel, leaving the container with two different kernels, thus the change

### DIFF
--- a/training/common/driver-toolkit/Containerfile
+++ b/training/common/driver-toolkit/Containerfile
@@ -7,7 +7,7 @@ ARG ENABLE_RT=''
 
 USER root
 
-RUN dnf -y update \
+RUN dnf -y update --exclude kernel* \
     && dnf info --installed kernel > /dev/null || dnf install -y kernel-core \
     && if [ "${KERNEL_VERSION}" == "" ]; then \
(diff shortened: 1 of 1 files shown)


training/common/driver-toolkit/Containerfile
```diff
@@ -7,8 +7,7 @@ ARG ENABLE_RT=''
 
 USER root
 
-RUN dnf -y update \
-    && dnf info --installed kernel > /dev/null || dnf install -y kernel-core \
+RUN dnf info --installed kernel > /dev/null || dnf install -y kernel-core \
```
rhatdan: How about dnf -y update --exclude kernel\*

javipolo: Nice idea, trying it now :)

javipolo: Works perfectly, I changed the code and the commit message :+1:

rhatdan (approved):
LGTM

PR #638, opened 2024-06-26
Add missing trailing backslash

javipolo:


rhatdan (approved):
Wow how was that missed.
LGTM

PR #645, opened 2024-06-27
Make help the default target for Makefile

javipolo:


rhatdan:
Should be fixed by https://github.com/containers/ai-lab-recipes/pull/642

PR #648, opened 2024-06-28
remove checkout with path so pathing is fixed

Gregory-Pereira:
Hopefully should fix: https://github.com/containers/ai-lab-recipes/actions/runs/9705287709/job/26787086103


Gregory-Pereira:
going to sudo merge because this is non-invasive + tests non-triggered on PR

PR #650, opened 2024-06-30
Workaround hardlinked /etc changes

cgwalters:
This (not tested locally) change should work around https://github.com/containers/bootc/issues/637

It's all tricky because you'll only again see this when using the `vfs` backend. The `overlay` backend always breaks hardlinks. For more, see e.g. https://github.com/ostreedev/ostree-rs-ext/pull/408


rhatdan (approved):
LGTM

PR #656, opened 2024-07-02
Pull instructlab image with auth, if provided

ralphbean:
Upstream, this image can be pulled unauthenticated, but in other environments a user might want to include an image that exists in some repository that requires authentication to pull.

The person building the image needs to provide `--secret=id=instructlab-nvidia-pull/.dockerconfigjson,src=instructlab-nvidia-pull/.dockerconfigjson` when building the image in order to make the secret available.


lmilbaum (approved):
LGTM

PR #659, opened 2024-07-04
Adding rhc for telemetry

tsorya:
Adding rhc in order to gather telemetry with insights client.
We need to be able to run rhc connect, should run only on rhel


Gregory-Pereira:
Can you fill me in on the reasoning around the hold? Just want to understand if there is a specific reason were waiting on this.

tsorya:
Had a problem with running the image, held for better testing (though had same problem with main branch)

rhatdan (approved):
LGTM

PR #660, opened 2024-07-07
fix testing-framework

sallyom:
remove trace steps to isolate the path issue, then fix the path issue

### DIFF
--- a/.github/workflows/testing_framework.yaml
+++ b/.github/workflows/testing_framework.yaml
@@ -53,51 +53,6 @@ jobs:
         with:
           python-version: '3.11'
 
-      - name: Install Python dependencies for otel trace generation
-        run: |
-          pip install --no-cache-dir opentelemetry-sdk opentelemetry-exporter-otlp opentelemetry-instrumentation
-
-      - name: Configure OpenTelemetry Collector
-        run: |
-          echo '
-          receivers:
-            otlp:
-              protocols:
-                grpc:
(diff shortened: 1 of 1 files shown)


sallyom:
@Gregory-Pereira @cooktheryan @lmilbaum ptal, thanks!

.github/workflows/testing_framework.yaml
```diff
@@ -53,50 +53,8 @@ jobs:
...
-          ./otelcol --config otel-collector-config.yaml &
-
-      - name: Start integration-tests trace
-        run: |
-          export WORKFLOW_NAME="testing_framework"
-          export STEP_NAME="integration-tests"
-          export TRACE_ACTION="start"
-          python ci/trace-steps.py
+      - name: Debug - List directory contents
+        run: ls -al ./recipes/natural_language_processing/chatbot
```
Gregory-Pereira: Do we still need this?

Gregory-Pereira (approved):
/lgtm, just reccomend removing 1 debug step but either way it passes muster, thanks

lmilbaum (approved):
LGTM

.github/workflows/testing_framework.yaml
sallyom: we don't need it, i'll remove it

sallyom:
@Gregory-Pereira I removed the debug statement and also re-commented the `on` section to run on PR

PR #661, opened 2024-07-08
Make sure ilab wrapper is executable and in path

javipolo:


training/nvidia-bootc/Containerfile
```diff
@@ -165,7 +165,9 @@ RUN grep -q /usr/lib/containers/storage /etc/containers/storage.conf || \
 	  cp /run/.input/ilab /usr/local/bin/ilab; \
      else \
 	  curl -o /usr/local/bin/ilab "https://raw.githubusercontent.com/containers/ai-lab-recipes/main/training/ilab-wrapper/ilab"; \
-    fi
+    fi \
+    && chmod +x /usr/local/bin/ilab \
+    && echo 'PATH=$PATH:/usr/local/bin' > /etc/profile.d/usrlocalbin.sh
```
Gregory-Pereira: I was under the assumption `/usr/local/bin` is in the default `$PATH` so when the OS is installed via the `bootc` container it should be there. Is this assumption incorrect?

javipolo: It is for users, but not for root, and IIRC ilab should be run as root

```
[cloud-user@javipolo-rhel-ai-0705a ~]$ echo $PATH
/var/home/cloud-user/.local/bin:/var/home/cloud-user/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin

[cloud-user@javipolo-rhel-ai-0705a ~]$ sudo -i
[root@javipolo-rhel-ai-0705a ~]# echo $PATH
/usr/local/sbin:/sbin:/bin:/usr/sbin:/usr/bin
```

rhatdan: Why not install ibal in /usr/bin/  ?usr/local should not be used.

javipolo: I just changed the commit to copy to `/usr/bin` and stop using /usr/local/bin

rhatdan (approved):
LGTM

PR #663, opened 2024-07-09
RHELAI-429: Adding upgrade informer service

tsorya:
Upgrade informer will run every couple of our and will be triggered by systemd timer.
In order to start it on boot and run once i enabled it and timer. Disabling auto upgrade service in order to remove unexpected reboots. Service will run "bootc upgrade --check" and in case new version exists it will create motd file with upgrade info.
Removed unused grow-part services


training/common/usr/lib/systemd/system/upgrade-informer.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Check for available bootc updates
```
jeremyeder: s/bootc/operating system

training/common/usr/lib/systemd/system/upgrade-informer.timer
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Runs upgrade informer periodically
+ConditionPathExists=/run/ostree-booted
+
+[Timer]
+OnBootSec=1h
+# This time is relatively arbitrary and obviously expected to be overridden/changed
```
jeremyeder: dont want to encourage people to change this, so please drop the comment. usage of OnBootSec is ok.

training/common/usr/libexec/upgrade-informer
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# Run the command and capture its output
+output=$(bootc upgrade --check | sed -e 1q)
+message_file="/etc/motd.d/upgrade-message"
+bootc_auth="/etc/ostree/auth.json"
+
+if [[ $output == Update\ available* ]]; then
```
jeremyeder, line 8: we need to get proper support into bootc so we dont have to parse strings.

training/common/usr/libexec/upgrade-informer
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# Run the command and capture its output
+output=$(bootc upgrade --check | sed -e 1q)
+message_file="/etc/motd.d/upgrade-message"
+bootc_auth="/etc/ostree/auth.json"
```
jeremyeder, line 6: how will this file get there? what happens if it doesnt exist?

tsorya: Currently we talked that user will need to add it.
We can run command like :
`podman login registry.redhat.io --username user --password pass --verbose | grep Used | awk '{print $2}' | xargs -I{} cp {} /etc/ostree/auth.json
`
There is a talk in bootc to add some common auth file between bootc and podman but it is still talks

training/common/usr/libexec/upgrade-informer
tsorya: Hmm, didn't find a good way to do it rather this one. I assume i can look on bootc and add , --json maybe?

tsorya: There is no good way to do it, bootc upgrade --check doesn't have an option for json format. 
Though there is draft pr for it  https://github.com/containers/bootc/pull/472#issuecomment-2061337390.
Hope we will be able to change it in future versions

rhatdan:
My understanding of the growpart was if the RHEL-AI was installed onto a cloud machine. Is this something we do not intend to do?

training/common/usr/libexec/upgrade-informer
rhatdan: Why don't we automatically download the image, and not reboot?

tsorya: We can run "bootc upgrade" without applying the change but it means that next reboot user will get new version in anycase, no? 
We talked about updating user that there is an available upgrade but not applying anything by ourselves for now at least

tsorya:
> My understanding of the growpart was if the RHEL-AI was installed onto a cloud machine. Is this something we do not intend to do?

@rhatdan it was removed from all the images so i assumed it was not needed anymore but wanted to use same process of adding new service. If it is required i can leave it

training/common/usr/libexec/upgrade-informer
tsorya: @jeremyeder WDYT?

training/common/usr/libexec/upgrade-informer
```diff
@@ -0,0 +1,32 @@
...
+    bootc_image=$(echo "$output" | awk '{print $4}')
+    # If auth file exists we should use it
+    auth_params=""
+    if [[ -f $bootc_auth ]]; then
+      auth_params="--authfile $bootc_auth"
+    fi
+
+    # Get image version
+    # shellcheck disable=SC2086
+    image_version=$(skopeo inspect --format json $auth_params "$bootc_image" | jq '.Labels | .["image_version"] // empty' | tr -d '"')
```
rhatdan: Not crazy about requiring skopeo, I believe this package could disappear and be replaced soley with Podman in the future.  Can't we just save the output of bootc upgrade --check?

@cgwalters PTAL

tsorya: We can though i wanted to get version from the image and bootc doesn't provide an option to get labels.
I assume we can change this part whenever we want. I think when we will remove skopeo this small service will not be there already and will be changed to something much mature.

tsorya: This is not must for sure and we can remove this part, we can update that there is new upgrade without saying which version, at least for now

training/common/usr/libexec/upgrade-informer
jeremyeder: showing folks and update is available is all we set out to do here; over time we can gauge what level of effort to put on this area.  we want to nudge users to upgrade for all the normal reasons, but going beyond that feels premature.

training/common/usr/libexec/upgrade-informer
jeremyeder: upgrade-informer would run before the user has a chance to authenticate unless you're collecting it in firstboot (which we talked about doing?)

tsorya: We can create doc how to add creds on boot but i am not sure how we can make customers to do it.
I can fail nice in case we can't reach registry and exit 0

tsorya: We can add systemd service that can collect username/password as env variables and connect system to registry + insights on boot? Will it be more helpful? 
I can block this service if /etc/ostree/auth.json was not set? Not sure what is better, not to run or try and fail gracefully in case no auth needed.

tsorya: What will happened with current code 
`Jul 15 13:23:50 localhost upgrade-informer[1645]: ERROR Upgrading: Creating importer: Failed to invoke skopeo proxy method OpenImage: remote error: reading manifest v3.15 in quay.io/itsoiref/nvidia-bootc-private: unauthorized: access to the requested re>
Jul 15 13:23:50 localhost upgrade-informer[1643]: No upgrade was found
`
It will not fail the service and just log the error
@jeremyeder i think it should be safe enough for now, WDYT?

training/common/usr/libexec/upgrade-informer
```diff
@@ -0,0 +1,32 @@
...
+
+    # If upgrade available, write the output to the file
+    echo -e "\n\n   **   Attention!   ** \n** A new $image_version version is available **\n\
```
tsorya: @jeremyeder can you help me with good text here?

javipolo:
This functionality has been splitted and moved into different PRs:

https://github.com/containers/ai-lab-recipes/pull/690
https://github.com/containers/ai-lab-recipes/pull/689

PR #665, opened 2024-07-10
RHELAI-473: Adding rhel ai version

tsorya:
Adding rhel ai version
Set github hash by defautl as image version.
Add RHEL AI as variant into /etc/os-release in order to use it in insights

NAME="Red Hat Enterprise Linux"
VERSION="9.20240630.0.4 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="9.4"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux 9.20240630.0.4 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_BUGZILLA_PRODUCT_VERSION=9.4
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.4"
OSTREE_VERSION='9.20240630.0'
VARIANT_ID=rhel_ai1
VARIANT="RHEL AI"
BUILD_ID='v1.1.3'

### DIFF
--- a/training/nvidia-bootc/Containerfile
+++ b/training/nvidia-bootc/Containerfile
@@ -99,6 +99,8 @@ COPY --from=builder --chmod=444 /home/builder/yum-packaging-precompiled-kmod/tmp
 # Temporary workaround until the permanent fix for libdnf is merged
 COPY nvidia-toolkit-firstboot.service /usr/lib/systemd/system/nvidia-toolkit-firstboot.service
 
+ARG IMAGE_VERSION
+
 # TODO: rework this monstrosity into a build.sh (or even not shell script)
 # The need for the `cp /etc/dnf/dnf.conf` is a workaround for https://github.com/containers/bootc/issues/637
 RUN mv /etc/selinux /etc/selinux.tmp \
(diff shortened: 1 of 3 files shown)


Gregory-Pereira:
This strikes me as a change we should add downstream as a new containerfile or a patch rather than here in the upstream, but not entirely sure. cc @fabiendupont @lmilbaum

Gregory-Pereira:
> This strikes me as a change we should add downstream as a new containerfile or a patch rather than here in the upstream, but not entirely sure. cc @fabiendupont @lmilbaum

Just found your comment in slack, the above is resolved by this but going to post it here for posterity: 

> We should set real version in downstream build, just wanted to have some default

fabiendupont:
The OS can be something else than RHEL, so not sure that `RHEL` in the variable name is relevant.
/cc @romfreiman @n1hility

tsorya:
@fabiendupont do we have some more generic name for it? 
I mean i can change to make it ${OS_ID}_AI_VERSION or just something hardcoded that will match better, just don't know what.

tsorya:
@Gregory-Pereira i think setting git hash in upstream or provide the way for dev to setup their own versioning can be helpful too and in downstream we will set release versions

rhatdan:
I agree this should be done downstream, not upstream.

tsorya:
> I agree this should be done downstream, not upstream.

Currently downstream uses upstream containerfile. I added verification that it is rhel image in order to set this field only for rhel

tsorya:
/unhold
Seems like we have decision

training/nvidia-bootc/Containerfile
```diff
@@ -146,6 +148,9 @@ RUN mv /etc/selinux /etc/selinux.tmp \
     # Install rhc connect for insights telemetry gathering
     && . /etc/os-release && if [ "${ID}" == "rhel" ]; then \
         dnf install -y rhc rhc-worker-playbook; \
+        sed -i -e "/^VARIANT=/ {s/^VARIANT=.*/VARIANT=\"RHEL AI\"/; t}" -e "\$aVARIANT=\"RHEL AI\"\"" /etc/os-release;
```
romfreiman, line 151: amd, intel?

tsorya: Will do in next PR, in anycase GA is nvidia.

lmilbaum:
Joining late to the party. What does this variable used for? If we agree that it is needed, should it be populated via a builds arg file?

lmilbaum (approved):
Approving this PR to unblock @tsorya from implementing the logic on the insights side. A greater effort is required which will probably change some of the bits contributed by this PR. Filing a Jira issue for the greater effort.

PR #666, opened 2024-07-11
[WIP] observability with bootc example (kepler & opentelemetry-collector)

sallyom:
This is PoC for adding observability (Kepler and OpenTelemetry Collector) to a bootc system. Not specific to AI workloads so not sure it belongs in this repo, but if everyone running AI workloads requires an observability solution, than perhaps it does belong here.

### DIFF
--- a/observability/kepler/otelcol-config.yaml
+++ b/observability/kepler/otelcol-config.yaml
@@ -1,0 +1,57 @@
+---
+extensions:
+  file_storage:
+    # this dir must exist on filesystem where otelcollector is running
+    directory: /etc/otelcol-logs
+    timeout: 10s
+receivers:
+  journald:
+    units:
+      - container-kepler
+      - opentelemetry-collector
+      #- ovn-kubernetes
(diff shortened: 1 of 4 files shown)


observability/kepler/otelcol-config.yaml
```diff
@@ -0,0 +1,57 @@
+---
+extensions:
+  file_storage:
+    # this dir must exist on filesystem where otelcollector is running
+    directory: /etc/otelcol-logs
```
frzifus, line 5: Isnt there a better place for cached data? AFAIK `/var` should be mutable? Maybe we can use something like `/var/cache/opentelemetry-collector/`? wdyt?

rhatdan: Yes lets treat /usr and /etc as read-only from apps point of view.

jeffmaury:
@sallyom more than 1 year old, can we close it ?

sallyom:
closing, stale

PR #668, opened 2024-07-11
Check for driver and kernel version mismatch

javipolo:
When a kernel module is installed via rpm, it installs kernel-core as a dependency

Only one kernel-core package should be in the final image, if there are two, most probably accelerator module was built for the wrong kernel version


Gregory-Pereira (approved):
/lgtm

PR #669, opened 2024-07-11
Fix SSH_PUBKEY expansion into build-arg

javipolo:
Since ssh public key has spaces, it was expanded multiple times


rhatdan (approved):
LGTM

PR #670, opened 2024-07-11
trigger konflux builds

Gregory-Pereira:
cc @lmilbaum @romfreiman 
should trigger builds


rhatdan:
This should not be merged correct?

Gregory-Pereira:
Yes, this can be closed now. Thanks!

PR #671, opened 2024-07-11
Add vendor label and arg to all containerfiles

ralphbean:
So that downstream vendors can indicate themselves on their builds.

I expect we will want to add more of [these](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys) in the future, but this one is a good start.


rhatdan:
Can you add a mechanism to set the Vendor from the Makefile.

rhatdan:
Not crazy about PROVIDER, but will merge.

What do you think about HARDWARE?

rhatdan (approved):
LGTM

ralphbean:
Yeah, `HARDWARE` works for me too.

PR #676, opened 2024-07-15
Testing membership to help for security reasons

cooktheryan:
Wanting to test a way to verify membership to the repo before tests will run

### DIFF
--- a/.github/workflows/training-e2e.yaml
+++ b/.github/workflows/training-e2e.yaml
@@ -27,6 +27,19 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
+  block-pr:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if PR is from forked repo
+        id: check_forked
+        run: |
+          if [[ ${{ github.event.pull_request.head.repo.full_name }} != ${{ github.event.pull_request.base.repo.full_name }} ]]; then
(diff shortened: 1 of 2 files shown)


cooktheryan:
@Gregory-Pereira @lmilbaum can you take a look at this approach?

.github/workflows/training-e2e.yaml
```diff
@@ -27,9 +27,48 @@ concurrency:
...
+
+  check-membership:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if user is a member of the organization
+        id: check_membership
+        run: |
+          ORG="your-organization"
+          USER=${{ github.actor }}
+          RESPONSE=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" "https://api.github.com/orgs/containers/members/$USER")
```
lmilbaum, line 51: Would that work if this pipeline triggered from a forked PR? Specifically asking about the secrets.GITHUB_TOKEN.

lmilbaum: Maybe there is a GitHub Actions predefined variable that has this value already populated

lmilbaum:
Would it make sense to develop this on a test repo and then apply it on a production pipeline?

.github/workflows/training-e2e.yaml
cooktheryan: so GITHUB_TOKEN should be a predefined variable defined in Github. So usually we do not set these.

PR #677, opened 2024-07-15
WIP: Fork access

cooktheryan:
This should allow for forks to use secrets but first committers should not have access to run actions based on repo permissions. Next, we verify the user is a member of the container organization. If not we fail the run which would stop them from getting secrets.


rhatdan:
@lmilbaum @Gregory-Pereira @sallyom PTAL

PR #682, opened 2024-07-18
RHELAI-524: Disable bootc auto upgrade service

tsorya:
We should not have autoupgrade process in rhel ai as we can't reboot in the middle of user workloads, we should disable bootc autoupgrade service


training/nvidia-bootc/Containerfile
```diff
@@ -154,6 +154,8 @@ RUN mv /etc/selinux /etc/selinux.tmp \
         sed -i -e "/^VARIANT=/ {s/^VARIANT=.*/VARIANT=\"RHEL AI\"/; t}" -e "\$aVARIANT=\"RHEL AI\"" /usr/lib/os-release; \
         sed -i -e "/^VARIANT_ID=/ {s/^VARIANT_ID=.*/VARIANT_ID=rhel_ai/; t}" -e "\$aVARIANT_ID=rhel_ai" /usr/lib/os-release; \
         sed -i -e "/^RHEL_AI_VERSION_ID=/ {s/^RHEL_AI_VERSION_ID=.*/RHEL_AI_VERSION_ID='${IMAGE_VERSION_ID}'/; t}" -e "\$aRHEL_AI_VERSION_ID='${IMAGE_VERSION_ID}'" /usr/lib/os-release; \
+        # disable auto upgrade service
+        rm -f /usr/lib/systemd/system/default.target.wants/bootc-fetch-apply-updates.timer; \
```
rhatdan, line 158: Please use systemctl disable.

tsorya: @rhatdan systemctl disable doesn't work. 
Service continues to be enabled, though removing simlink does the job. 
I assume this is the reason for other lines with adding simlinks to enable services in this file

rhatdan: So systemctl disable does not remove the link?

rhatdan:
Merging but I think systemctl disable should work.

PR #684, opened 2024-07-19
add skopeo so that ilab model download will work

jeremyeder:
This is required for https://github.com/instructlab/instructlab/pull/1756 to work.

PR #688, opened 2024-07-22
add skopeo so that ilab model download will work

rhatdan:
While skopeo maybe part of the base image, there is no guarantee, and as long as ilab requires it, we should make sure it is installed.

Replaces: https://github.com/containers/ai-lab-recipes/pull/684


rhatdan:
@jeremyeder @n1hility @Gregory-Pereira @cooktheryan PTAL

sallyom:
LGTM

sallyom (approved):

PR #689, opened 2024-07-23
RHELAI-429: Adding upgrade informer service

javipolo:
This is a duplication of https://github.com/containers/ai-lab-recipes/pull/663, since @tsorya is on PTO and we really need to push those changes

/cc @rhatdan 
/cc @jeremyeder

Upgrade informer will run every couple of our and will be triggered by systemd timer.

In order to start it on boot and run once i enabled it and timer. Disabling auto upgrade service in order to remove unexpected reboots. Service will run "bootc upgrade --check" and in case new version exists it will create motd file with upgrade info.

### DIFF
--- a/training/common/usr/lib/systemd/system/upgrade-informer.service
+++ b/training/common/usr/lib/systemd/system/upgrade-informer.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Check for available RHEL AI upgrade
+ConditionPathExists=/run/ostree-booted
+After=network-online.target
+StartLimitIntervalSec=400
+StartLimitBurst=3
+
+[Service]
+Type=oneshot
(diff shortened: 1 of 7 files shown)


training/common/usr/lib/systemd/system/upgrade-informer.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Check for available operating system updates
```
romfreiman: s/operating system/RHEL AI/ upgrade ?

rhatdan:
This should run only once per day.  If you run it more often then that you will see complaints about `flooding the logs`.

We are not going to be updating that often that it requires more then once/day to run.

javipolo:
> This should run only once per day

Agree, changed the timer to `1day`

PR #690, opened 2024-07-23
RHEL-33208: Remove custom bootc-generic-growpart

javipolo:
Since this is already fixed upstream, we don't need to add this custom service to grow filesystem


rhatdan:
LGTM

PR #691, opened 2024-07-24
fix the photo upload issue in object detection app

yangcao77:
This PR is to fix the issue while uploading a photo:
```
cannot write mode RGBA as JPEG
```

JPG does not support transparency - RGBA means Red, Green, Blue, Alpha - Alpha is transparency.

The Image class has a method convert which can be used to convert RGBA to RGB - after that it will be able to save as JPG.


yangcao77:
@MichaelClifford @rhatdan @Gregory-Pereira 
May I get a review on this PR? Thanks!

rhatdan:
LGTM

PR #695, opened 2024-07-25
Add support for changing Vendor and Buildhost for the driver RPM

kwozyman:


rhatdan:
LGTM

rhatdan:
Should we modify the Makefile to pass these in as well?  Is this a change we want for amd and intel?

kwozyman:
> Should we modify the Makefile to pass these in as well? Is this a change we want for amd and intel?

This change is specific to rpm building and currently we're not building packages for Intel and Amd.

PR #696, opened 2024-07-25
nvidia-bootc: Duplicate common lib & ilab, to avoid an intermediate makefile target

pastequo:
Following a discussion with @ralphbean , it would be preferable to avoid any makefile command before building this image in Konflux.

This PR dupplicates the required files

### DIFF
--- a/training/nvidia-bootc/Containerfile
+++ b/training/nvidia-bootc/Containerfile
@@ -104,7 +104,7 @@ COPY --from=builder --chmod=444 /home/builder/yum-packaging-precompiled-kmod/tmp
 # Temporary workaround until the permanent fix for libdnf is merged
 COPY nvidia-toolkit-firstboot.service /usr/lib/systemd/system/nvidia-toolkit-firstboot.service
 # Enable common services
-COPY build/usr /usr
+COPY duplicated/common/usr /usr
 
 ARG IMAGE_VERSION_ID
 
@@ -187,13 +187,10 @@ fi
 # Also make sure not to duplicate if a base image already has it specified.
 RUN grep -q /usr/lib/containers/storage /etc/containers/storage.conf || \
     sed -i -e '/additionalimage.*/a "/usr/lib/containers/storage",' \
-	/etc/containers/storage.conf && \
-    if [ -f "/run/.input/ilab" ]; then \
-	  cp /run/.input/ilab /usr/bin/ilab; \
-     else \
-	  curl -o /usr/bin/ilab "https://raw.githubusercontent.com/containers/ai-lab-recipes/main/training/ilab-wrapper/ilab"; \
-    fi \
-    && chmod +x /usr/bin/ilab
+	/etc/containers/storage.conf
+
+COPY duplicated/ilab-wrapper/ilab /usr/bin/ilab
+RUN chmod +x /usr/bin/ilab
 
 ARG INSTRUCTLAB_IMAGE="quay.io/ai-lab/instructlab-nvidia:latest"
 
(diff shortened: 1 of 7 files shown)


romfreiman:
@rhatdan ptal

training/nvidia-bootc/Containerfile
```diff
@@ -182,13 +182,10 @@ fi
 # Also make sure not to duplicate if a base image already has it specified.
 RUN grep -q /usr/lib/containers/storage /etc/containers/storage.conf || \
     sed -i -e '/additionalimage.*/a "/usr/lib/containers/storage",' \
-	/etc/containers/storage.conf && \
-    if [ -f "/run/.input/ilab" ]; then \
-	  cp /run/.input/ilab /usr/bin/ilab; \
-     else \
-	  curl -o /usr/bin/ilab "https://raw.githubusercontent.com/containers/ai-lab-recipes/main/training/ilab-wrapper/ilab"; \
```
pastequo, line 194: I though it was better to always pick the local version, to have build reproducibility

training/nvidia-bootc/Makefile
```diff
@@ -30,4 +30,4 @@ bootc: driver-toolkit check-sshkey prepare-files
 		--security-opt label=disable \
 		--tag "${BOOTC_IMAGE}" \
 		-v ${OUTDIR}:/run/.input:ro \
-		${CONTAINER_TOOL_EXTRA_ARGS} .
+		${CONTAINER_TOOL_EXTRA_ARGS} ..
```
rhatdan: Does this work if called from the makefile in the parent directory?

training/nvidia-bootc/Containerfile
```diff
@@ -203,12 +200,12 @@ VOLUME /var/lib/containers
 
 RUN --mount=type=secret,id=instructlab-nvidia-pull/.dockerconfigjson \
     if [ -f "/run/.input/instructlab-nvidia/oci-layout" ]; then \
-         IID=$(podman --root /usr/lib/containers/storage pull oci:/run/.input/instructlab-nvidia) && \
-         podman --root /usr/lib/containers/storage image tag ${IID} ${INSTRUCTLAB_IMAGE}; \
+        IID=$(podman --root /usr/lib/containers/storage pull oci:/run/.input/instructlab-nvidia) && \
+        podman --root /usr/lib/containers/storage image tag ${IID} ${INSTRUCTLAB_IMAGE}; \
     elif [ -f "/run/secrets/instructlab-nvidia-pull/.dockerconfigjson" ]; then \
-         IID=$(sudo podman --root /usr/lib/containers/storage pull --authfile /run/secrets/instructlab-nvidia-pull/.dockerconfigjson ${INSTRUCTLAB_IMAGE}); \
```
pastequo: Not sure why we put this in IID (not used in this case), or why some podman commands are launched with sudo and some w/o...

training/nvidia-bootc/Makefile
pastequo: like `make ./nvidia-bootc/Makefile bootc` ? If so it will fail early because of the Makefile inclusion

edit: didn't see the parent Makefile

romfreiman:
/hold
what are the minimum required changes to add the upgrade notification service?

training/nvidia-bootc/Makefile
rhatdan: More of `make nvidia` in the training directory.

pastequo:
@romfreiman New version, dupplicating lib & ilab

/cc @kwozyman @javipolo

javipolo:
LGTM

javipolo (approved):

Gregory-Pereira:
Can we fix DCO?

PR #697, opened 2024-07-25
ilab wrapper: `exec` for better signal handling / termination

omertuc:
If the wrapper script is killed, the container will be left running. Instead of just running the command, use `exec` to replace the wrapper script with the command, so that the command will receive the same signals as the wrapper script and the container will be terminated as expected.

### DIFF
--- a/training/ilab-wrapper/ilab
+++ b/training/ilab-wrapper/ilab
@@ -25,4 +25,4 @@ PODMAN_COMMAND=("podman" "run" "--rm" "-it"
 	"--env" "HF_TOKEN"
 	"${IMAGE_NAME}")
 
-"${PODMAN_COMMAND[@]}" "${PARAMS[@]}"
+exec "${PODMAN_COMMAND[@]}" "${PARAMS[@]}"


eranco74 (commented):
/lgtm
Thanks!!

eranco74:
I just checked and it works as expected.
I execute `ilab model serve` from systemd service, prior to this change if I execute `systemctl stop ilab.service` the ilab container would keep running  and the service journal log would say:
 ```
 Jul 25 12:46:40 ip-172-31-6-97.ec2.internal systemd[1672]: ilab.service: State 'final-sigterm' timed out. Killing.
Jul 25 12:46:40 ip-172-31-6-97.ec2.internal systemd[1672]: ilab.service: Killing process 75939 (conmon) with signal SIGKILL.
Jul 25 12:46:40 ip-172-31-6-97.ec2.internal systemd[1672]: ilab.service: Failed with result 'timeout'.
Jul 25 12:46:40 ip-172-31-6-97.ec2.internal systemd[1672]: Stopped ilab model serve service.
```
With this change the systemd service behaves properly.

rhatdan:
LGTM

PR #698, opened 2024-07-25
Use custom systemd files and ilab wrapper from inside nvidia-bootc build context

javipolo:
In order to addres a problem with the build pipeline, we need all the files to be present inside of each container build context

Until we switch to have .. as the build context, we're duplicating some files


javipolo:
Duplicates https://github.com/containers/ai-lab-recipes/pull/696 :facepalm:

PR #699, opened 2024-07-25
feat(nvidia-bootc): Introduce INSTRUCTLAB_IMAGE_PULL_SECRET

pastequo:
To be able to customize Konflux build with build-args. 

This is needed to keep the secret name sync-ed with whta is defined in Konflux

/cc @ralphbean


rhatdan:
LGTM

PR #700, opened 2024-07-25
Split main targets into instructlab + bootc target

javipolo:
In case we want to override `INSTRUCTLAB_IMAGE`, we should be able to skip the instructlab building process by just running:
```
make INSTRUCTLAB_IMAGE=blahblahblah nvidia-bootc
```


rhatdan:
LGTM

PR #704, opened 2024-07-29
fix(modell_server): deprecated make arguments for llamacpp server

axel7083:
The `LLAMA_CUBLAS` argument is deprecated

https://github.com/ggerganov/llama.cpp/blob/be20e7f49d9e5c6d9e8d9b4871eeba3df7a1639d/Makefile#L71-L72

### DIFF
--- a/model_servers/llamacpp_python/cuda/Containerfile
+++ b/model_servers/llamacpp_python/cuda/Containerfile
@@ -4,7 +4,7 @@ RUN  dnf install -y gcc-toolset-13-gcc gcc-toolset-13-gcc-c++
 USER 1001
 WORKDIR /locallm
 COPY src .
-ENV CMAKE_ARGS="-DLLAMA_CUBLAS=on -DLLAMA_AVX2=OFF -DLLAMA_FMA=OFF -DLLAMA_F16C=OFF"
+ENV CMAKE_ARGS="-DGGML_CUDA=1 -DLLAMA_AVX2=OFF -DLLAMA_FMA=OFF -DLLAMA_F16C=OFF"
 ENV FORCE_CMAKE=1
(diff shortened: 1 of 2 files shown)


lmilbaum:
Would it make more sense to update the Renovate PR https://github.com/containers/ai-lab-recipes/pull/703 instead?

axel7083:
> Would it make more sense to update the Renovate PR #703 instead?

Hi @lmilbaum there are two elements in this PR,
- First the fix for the containerfile as mention in the description `The LLAMA_CUBLAS argument is deprecated`, 
- Second the bump of libraries

I did not see https://github.com/containers/ai-lab-recipes/pull/703 when I opened this PR this morning, thanks for informing me. I will be removing the bump of library to keep the focus on the fix.

Gregory-Pereira:
Shouldn't this be:
```
CMAKE_ARGS="-DGGML_CUDA=on"
```
[according to the docs](https://github.com/abetlen/llama-cpp-python/tree/v0.2.82?tab=readme-ov-file#supported-backends). Guess im confused why we are turning it off when this variant is specifically cuda.

model_servers/llamacpp_python/cuda/Containerfile
```diff
@@ -4,7 +4,7 @@ RUN  dnf install -y gcc-toolset-13-gcc gcc-toolset-13-gcc-c++
 USER 1001
 WORKDIR /locallm
 COPY src .
-ENV CMAKE_ARGS="-DLLAMA_CUBLAS=on -DLLAMA_AVX2=OFF -DLLAMA_FMA=OFF -DLLAMA_F16C=OFF"
+ENV CMAKE_ARGS="-DGGML_CUDA=1 -DLLAMA_AVX2=OFF -DLLAMA_FMA=OFF -DLLAMA_F16C=OFF"
```
axel7083:
```suggestion
ENV CMAKE_ARGS="-DGGML_CUDA=on -DLLAMA_AVX2=OFF -DLLAMA_FMA=OFF -DLLAMA_F16C=OFF"
```

rhatdan (approved):
LGTM

PR #707, opened 2024-07-30
ilab wrapper: fix formatting

omertuc:
The wrapper had a mixed used of tabs/spaces, making it annoying to edit

Formatted with shfmt to switch to spaces


rhatdan (approved):
LGTM

PR #709, opened 2024-07-31
RHELAI-744 - Use environment variable to override HOME in ilab wrapper

javipolo:


rhatdan:
Why wouldn't I just override HOME?

HOME=/tmp/foobar ilab ...

fabiendupont:
The `ILAB_HOME` variable can be persistent (`.bash_profile`...), so that the user doesn't have to type `HOME=/tmp/foobar` on every command line. And alias would also work, but a convention on an environment variable name is easier to override for tests.

rhatdan:
LGTM

PR #713, opened 2024-08-01
ilab-wrapper: Run podman with sudo

omertuc:
Solves RHELAI-740

# Background

The ilab command is wrapped by an `ilab` script which launches ilab inside a podman container.

# Issue

Since the ilab container image is pulled during the bootc image build process using the root user, the image is not accessible to non-root users.

# Solution

We run the container as sudo in order to be able to access the root container storage. But for security reasons we map root UID 0 inside the container to the current user's UID (and all the other subuids to the user's /etc/subuid range) so that we're effectively running the container as the current user.

# Additional changes

Changed `"--env" "HOME"` to `"--env" "HOME=$HOME"` to pass the HOME environment variable from the current shell and not from the sudo environment.

# Future work

In the future, we will run podman as the current user, once we figure a reasonable way for the current user to access the root's user container storage

### DIFF
--- a/training/ilab-wrapper/ilab
+++ b/training/ilab-wrapper/ilab
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+function echo-err { echo "$@" >&2; }
+
 # Template values replaced by container build
 CONTAINER_DEVICE="__REPLACE_CONTAINER_DEVICE__"
 IMAGE_NAME="__REPLACE_IMAGE_NAME__"
@@ -8,7 +10,7 @@ export ENTRYPOINT="/opt/python3.11/venv/bin/ilab"
 export PARAMS=("$@")
 
 if [[ -n "$ILAB_HOME" ]]; then
-    HOME="$ILAB_HOME"
+	HOME="$ILAB_HOME"
 fi
 
 for dir in "$HOME/.cache" "$HOME/.config" "$HOME/.local"; do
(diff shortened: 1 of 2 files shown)


training/ilab-wrapper/ilab
```diff
@@ -8,21 +8,39 @@ export ENTRYPOINT="/opt/python3.11/venv/bin/ilab"
...
+# storage, which has the ilab image pre-pulled. But for security reasons we map
+# root UID 0 inside the container to the current user's UID (and all the other
+# subuids to the user's /etc/subuid range) so that we're effectively running
+# the container as the current user.
+#
+# In the future, we will run podman as the current user, once we figure a
+# reasonable way for the current user to access the root's user container
+# storage.
+CURRENT_USER_NAME=$(id --user --name)
+CURRENT_USER_SUBUID_RANGE=$(awk \
```
rhatdan, line 55: Should check UID as well as username, both are valid.

rhatdan: There can be multiple lines also, but for now, keep it simple.

This should really be provided via getent as well, but it does not look like it is currently.

omertuc: >There can be multiple lines also, but for now, keep it simple.

As in, multiple lines could match the same user/uid?

omertuc: If yes, added hard error for that scenario, for now

training/ilab-wrapper/ilab
```diff
@@ -8,21 +8,54 @@ export ENTRYPOINT="/opt/python3.11/venv/bin/ilab"
 export PARAMS=("$@")
 
 for dir in "$HOME/.cache" "$HOME/.config" "$HOME/.local"; do
```
rhatdan, line 16: Should probably check if $HOME is set. Rather then creating /.cache and /.config

omertuc: Not sure we even still need this if statement. Will address in a separate PR

training/ilab-wrapper/ilab
```diff
@@ -8,21 +8,54 @@ export ENTRYPOINT="/opt/python3.11/venv/bin/ilab"
...
+CURRENT_USER_NAME=$(id --user --name)
+CURRENT_USER_SUBUID_RANGE=$(awk \
+	--field-separator ':' \
+	--assign current_user="$CURRENT_USER_NAME" \
+	--assign current_uid="$UID" \
+	'$1 == current_user || $1 == current_uid {print $2 ":" $3}' \
+	/etc/subuid)
+
+# TODO: Handle multiple subuid ranges, for now, hard fail
+if [[ $(echo "$CURRENT_USER_SUBUID_RANGE" | wc -l) != 1 ]]; then
```
rhatdan:
```
$(wc -l <<< ${CURRENT_USER_SUBUID_RANGE})
```

training/ilab-wrapper/ilab
```diff
@@ -8,21 +8,54 @@ export ENTRYPOINT="/opt/python3.11/venv/bin/ilab"
...
+	if [[ -z "$CURRENT_USER_SUBUID_RANGE" ]]; then
+		echo "No subuid range found for user $CURRENT_USER_NAME ($UID)"
+	else
+		echo "Multiple subuid ranges found for user $CURRENT_USER_NAME ($UID):"
```
rhatdan: Might want to use printf rather then echo.

training/ilab-wrapper/ilab
```diff
@@ -8,21 +8,54 @@ export ENTRYPOINT="/opt/python3.11/venv/bin/ilab"
...
+	--field-separator ':' \
+	--assign current_user="$CURRENT_USER_NAME" \
+	--assign current_uid="$UID" \
+	'$1 == current_user || $1 == current_uid {print $2 ":" $3}' \
+	/etc/subuid)
+
+# TODO: Handle multiple subuid ranges, for now, hard fail
+if [[ $(echo "$CURRENT_USER_SUBUID_RANGE" | wc -l) != 1 ]]; then
+	if [[ -z "$CURRENT_USER_SUBUID_RANGE" ]]; then
+		echo "No subuid range found for user $CURRENT_USER_NAME ($UID)"
```
rhatdan: I think  you should add a function like 
function echo-err { echo "$@" >&2; }

And then instead of using echo  use echo-err to make sure error messages go to stderr.

training/ilab-wrapper/ilab
rhatdan: Similar to above this should go to stderr

training/nvidia-bootc/duplicated/ilab-wrapper/ilab
```diff
@@ -8,21 +8,54 @@ export ENTRYPOINT="/opt/python3.11/venv/bin/ilab"
 export PARAMS=("$@")
 
 for dir in "$HOME/.cache" "$HOME/.config" "$HOME/.local"; do
```
rhatdan, line 16: Why is this duplicated?

training/ilab-wrapper/ilab
omertuc: #717

training/nvidia-bootc/duplicated/ilab-wrapper/ilab
omertuc: See https://github.com/containers/ai-lab-recipes/pull/708#issuecomment-2258421931

rhatdan:
Needs rebase.

omertuc:
Force pushed for rebase & resolving review comments

training/ilab-wrapper/ilab
omertuc: Any particular reason you think we should use `printf`?

rhatdan: Makes writing multi line messages easier.  But It is not a big deal.

rhatdan:
LGTM

training/ilab-wrapper/ilab
omertuc: FYI this doesn't work because Here-Strings always have a newline appended to them

PR #716, opened 2024-08-01
Set group permissions on the /rag/models/ dir

maysunfaisal:
Updates the RAG Containerfile for setting group permissions on the /rag/models/ dir.

We have been consuming the Podman Desktop AI lab recipes via RHDH [software templates](https://github.com/redhat-ai-dev/ai-lab-template/tree/main/templates), which will include OpenShift as a deployment target for the recipes. While trying to run the RAG application, we found it was running into permissions issues when using the UID that OpenShift assigns to the container. The fix has been verified both locally and on OpenShift.

Previously on OpenShift:
Screenshot 2024-08-01 at 6 13 47 PM


cooktheryan:
This is awesome but will defer to @rhatdan to decide

rhatdan (approved):
LGTM

PR #718, opened 2024-08-02
Require KERNEL_VERSION argument in driver-toolkit

fabiendupont:
The default base image for the Driver Toolkit image is `centos:stream9`. The original work for Driver Toolkit is in OpenShift and the base image is `ubi9/ubi`. In bother cases, the images don't have the `kernel` package installed.

This change adds a test on the `KERNEL_VERSION` argument and exits if it's not provided at build time. This also ensure that only the relevant kernel is present when using `centos:stream9` or `ubi9/ubi` as the base image. And this realigns a bit with the original Driver Toolkit.

### DIFF
--- a/training/common/driver-toolkit/Containerfile
+++ b/training/common/driver-toolkit/Containerfile
@@ -11,26 +11,25 @@ ARG ENABLE_RT=''
 
 USER root
 
-# TODO: rework this monstrosity into a build.sh (or even not shell script)
 # The need for the `cp /etc/dnf/dnf.conf` is a workaround for https://github.com/containers/bootc/issues/637
-RUN mv /etc/selinux /etc/selinux.tmp \
-    && dnf -y update --exclude kernel* \
-    && dnf info --installed kernel > /dev/null || dnf install -y kernel-core \
-    && if [ "${KERNEL_VERSION}" == "" ]; then \
-        RELEASE=$(dnf info --installed kernel-core | awk -F: '/^Release/{print $2}' | tr -d '[:blank:]') \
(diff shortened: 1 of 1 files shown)


training/common/driver-toolkit/Containerfile
```diff
@@ -11,26 +11,25 @@ ARG ENABLE_RT=''
...
-RUN mv /etc/selinux /etc/selinux.tmp \
-    && dnf -y update --exclude kernel* \
-    && dnf info --installed kernel > /dev/null || dnf install -y kernel-core \
-    && if [ "${KERNEL_VERSION}" == "" ]; then \
-        RELEASE=$(dnf info --installed kernel-core | awk -F: '/^Release/{print $2}' | tr -d '[:blank:]') \
-        && VERSION=$(dnf info --installed kernel-core | awk -F: '/^Version/{print $2}' | tr -d '[:blank:]') \
-        && export KERNEL_VERSION="${VERSION}-${RELEASE}" ;\
-        fi \
-    && echo "${KERNEL_VERSION}" \
+RUN if [ "${KERNEL_VERSION}" == "" ] ; then \
```
rhatdan: I prefer  
test -z ${KERNEL_VERSION}

rhatdan: If this is not a bootable container image ,is there a reason for requiring the kernel?

fabiendupont: The Driver Toolkit is only used to compile out-of-tree kernel modules. It is used for the `intel-bootc` and `nvidia-bootc` image build with a multi-stage approach. The drivers are built in DriverToolkit, then copied over the final image, which is bootable.

rhatdan:
LGTM

PR #722, opened 2024-08-04
Add environment variables for ilab container and increase shm size for vllm

### LINKED ISSUE #721: Need to increase shm size for ilab launcher script
Currently: the ilab wrapper script here: https://github.com/containers/ai-lab-recipes/blob/main/training/ilab-wrapper/ilab does not update the shm size to be 10 Gigs. It is noted in requirements for running deepspeed and vllm that a shm size of 10GB is necessary to run full scale model inferencing and training. 

A related issue: https://github.com/vllm-project/vllm/issues/1710 and just to note: the earlier scripts that launched vllm directly had a shm size of 10GB

relyt0925:
Include ILAB_GLOBAL_CONFIG, VLLM_LOGGING_LEVEL, and NCCL_DEBUG as environment variables when starting the ilab container. Also add shared memory size of 10G to enable vllm execution. Resolves: https://github.com/containers/ai-lab-recipes/issues/721


rhatdan:
LGTM

rhatdan (approved):

PR #723, opened 2024-08-04
Comply with bashate style rules and remove exports

javipolo:
- Set all indenting to 4 spaces (no tabs)
- Use POSIX style function definition in oneliner functions
- Remove unneeded exports on env variables

/cc @omertuc


rhatdan:
@n1hility PTAL

PR #725, opened 2024-08-05
feat(KONFLUX-3933): Use rpm to get the value

jhutar:
This might be a more straight-forward way how to get a kernel version and release.

Initially I thought there will be a run duration benefit (running whole `dnf` vs. just `rpm`), but that is not a case (at least in this `builder` container).


rhatdan:
You have to sign your commits.

git commit -a --amend -s
git push --force

jhutar:
Thank you! Done.

rhatdan:
LGTM

PR #729, opened 2024-08-06
ilab-wrapper: don't map UIDs if we're already running as root

omertuc:
# Background

See df8885777d2257a6271ca1a6461673d1d96e99d6

# Issue

Introduced a regression [1] where it's no longer possible to run the script as root, as the subuid map ends up being empty and this causes an error:

```
Error: invalid empty host id at UID map: [1  1]
```

# Solution

Avoid UID mapping if we're already running as root.

# Motivation

We want to also be able to run the script as root, for example as part of a systemd service.

[1] RHELAI-798

### DIFF
--- a/training/ilab-wrapper/ilab
+++ b/training/ilab-wrapper/ilab
@@ -60,23 +60,30 @@ fi
 # In the future, we will run podman as the current user, once we figure a
 # reasonable way for the current user to access the root's user container
 # storage.
-CURRENT_USER_NAME=$(id --user --name)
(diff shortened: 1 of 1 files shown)


omertuc:
cc @eranco74

omertuc:
Force push reason: DCO

rhatdan:
LGTM

omertuc:
Forgot to update duplicated wrapper, followup (& additional fixes) here: https://github.com/containers/ai-lab-recipes/pull/730

PR #732, opened 2024-08-07
hooks: pre-commit hook to duplicate file

omertuc:
We have a file that's always a duplicate of another file, until we can get rid of this requirement a pre-commit hook to take care of it would be nice

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,3 +56,15 @@ currently built images are  tracked in
 ## [Training](./training/README.md)
 
 Linux Operating System Bootable containers enabled for AI Training
+
+## Setting Up Git Hooks
+
+To install our standard git hooks, run the following command:
+
+```sh
+./install-hooks.sh
+```
+
+### pre-commit hook
+
+Ensures that `training/ilab-wrapper/ilab` is duplicated into `training/nvidia-bootc/duplicated/ilab-wrapper/ilab`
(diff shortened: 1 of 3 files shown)


training/nvidia-bootc/duplicated/ilab-wrapper/ilab
```diff
@@ -60,23 +60,30 @@ fi
...
 # reasonable way for the current user to access the root's user container
 # storage.
-CURRENT_USER_NAME=$(id --user --name)
-CURRENT_USER_SUBUID_RANGE=$(awk \
-    --field-separator ':' \
-    --assign current_user="$CURRENT_USER_NAME" \
-    --assign current_uid="$UID" \
-    '$1 == current_user || $1 == current_uid {print $2 ":" $3}' \
-    /etc/subuid)
+if [[ "$UID" == 0 ]]; then
```
rhatdan: This should not be part of this PR.

omertuc: I disagree, it's syncing the duplicated wrapper as expected from the hook

omertuc: But looks like #731 already did that

rhatdan (approved):
LGTM

PR #733, opened 2024-08-07
Fix NVIDIA FabricManager unit file

fabiendupont:
In the `nvidia-bootc` Containerfile, the condition on the existence of `/dev/nvswitchctl` in the `nvidia-fabricmanager` unit file is not persisted, because we don't use the `-i` option of `sed`, so the final image still always tries to load the service. This change adds the `-i` option to fix this.


Gregory-Pereira (approved):
/lgtm

PR #735, opened 2024-08-07
Fix the retrieval of kernel version, release and dist

fabiendupont:
During the build of the out-of-tree drivers, the base image will always have the `kernel-core` package installed. And the `Makefile` doesn't pass the `KERNEL_VERSION` argument to the build command. So, it's simpler to rely on the `kernel-core` package info.

The commands to get the `KREL` and `KDIST` were not working with RHEL 9.4 kernel. The new set of commands has been tested with `ubi9/ubi:9.4` and `centos/centos:stream9` based driver toolkit image and they return the correct value. For example, the values returned for the following kernels are:

* `5.14.0-427.28.1.el9_4` (`ubi9:ubi:9.4`):
 * `KVER`: `5.14.0`
 * `KREL`: `427.28.1`
 * `KDIST`: `.el9_4`
* `5.14.0-427.el9` (`centos/centos:stream9`):
 * `KVER`: `5.14.0`
 * `KREL`: `427`
 * `KDIST`: `.el9`

The `OS_VERSION_MAJOR` argument is also not passed by the `Makefile`, but we can get it from the `/etc/os-release` file. I'm switching to grep+sed, because I don't want to load all the other variables.

### DIFF
--- a/training/nvidia-bootc/Containerfile
+++ b/training/nvidia-bootc/Containerfile
@@ -6,9 +6,6 @@ FROM ${DRIVER_TOOLKIT_IMAGE} as builder
 
 ARG BASE_URL='https://us.download.nvidia.com/tesla'
 
-ARG OS_VERSION_MAJOR=''
-ARG KERNEL_VERSION=''
-
 ARG BUILD_ARCH=''
 ARG TARGET_ARCH=''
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

rhatdan (approved):

PR #736, opened 2024-08-07
Driver Toolkit: Get kernel version from bootc image

fabiendupont:
Driver Toolkit: Get kernel version from bootc image

When building the `driver-toolkit` image, It is cumbersome to find kernel
version that matches the future `nvidia-bootc` and `intel-bootc` images.
However, the kernel version is stored as a label on the `rhel-bootc`
images, which are exposed as the `FROM` variable in the Makefile.

This change collects the kernel version using `skopeo inspect` and `jq`.

The `DRIVER_TOOLKIT_BASE_IMAGE` variable is introduced in the Makefile
to dissociate it from the `FROM` variable that is used as the `nvidia-bootc`
and `intel-bootc` base image.

The user can now specify something like:

```shell
make nvidia-bootc \
    FROM=quay.io/centos-bootc/centos-bootc:stream9 \
    DRIVER_TOOLKIT_BASE_IMAGE=quay.io/centos/centos:stream9
```

Also, the `VERSION` variable in `/etc/os-release` is the full version, so
this change modifies the command to retrieve the `OS_VERSION_MAJOR`
value.

Signed-off-by: Fabien Dupont <fdupont@redhat.com>

### DIFF
--- a/training/common/Makefile.common
+++ b/training/common/Makefile.common
@@ -31,10 +31,10 @@ DISK_TYPE ?= qcow2
 DISK_UID ?= $(shell id -u)
 DISK_GID ?= $(shell id -g)
 
-ARCH ?=
+ARCH ?= $(shell arch)
 
 DRIVER_VERSION ?=
-KERNEL_VERSION ?=
+KERNEL_VERSION ?= $(shell skopeo inspect --format json docker://${FROM} | jq -r '.Labels["ostree.linux"]' | sed "s/\.${ARCH}//")
 
 INSTRUCTLAB_IMAGE ?= $(REGISTRY)/$(REGISTRY_ORG)/instructlab-$(HARDWARE):$(IMAGE_TAG)
 WRAPPER = $(CURDIR)/../ilab-wrapper/ilab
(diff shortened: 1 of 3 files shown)


training/common/Makefile.common
```diff
@@ -113,7 +114,7 @@ driver-toolkit:
 		$(ARCH:%=--platform linux/%) \
 		$(BUILD_ARG_FILE:%=--build-arg-file=%) \
 		$(ENABLE_RT:%=--build-arg ENABLE_RC=%) \
-		$(FROM:%=--from=%) \
+		$(DRIVER_TOOLKIT_BASE_IMAGE:%=--build-arg BASE_IMAGE=%) \
```
javipolo: Let's use `BASEIMAGE` to mantain coherence with other Containerfiles

training/common/driver-toolkit/Containerfile
```diff
@@ -1,6 +1,6 @@
-ARG BASEIMAGE="quay.io/centos/centos:stream9"
```
javipolo: Let's use BASEIMAGE to mantain coherence with other Containerfiles

training/common/driver-toolkit/Containerfile
```diff
@@ -1,6 +1,6 @@
...
+ARG BASE_IMAGE="quay.io/centos/centos:stream9"
 
-FROM ${BASEIMAGE}
```
javipolo: Let's use BASEIMAGE to mantain coherence with other Containerfiles

javipolo (changes requested):

training/common/Makefile.common
fabiendupont: I reverted. But we're using `_` to separate words in environment variables, so it looked more consistent with the other variables. Maybe, I'll propose another PR to use `BASE_IMAGE` everywhere 😉

lmilbaum (approved):

javipolo (approved):

PR #737, opened 2024-08-07
Add extra message on ilab error with docs outside the VPN

javipolo:


javipolo:
Let's hold this until there is consensus on using this approach :sweat_smile:

javipolo:
We need to create a more thorough PR with all the implementation of this

PR #739, opened 2024-08-07
ilab-wrapper: support execution by non-sudoers

omertuc:
DRAFT because I didn't test the `Containerfile` changes yet.

Also, the way it's currently implemented, I guess there's nothing stopping the user from abusing this script to run pretty much any command.... We're gonna have to think of a way to lock it down so the parameters constructed by the pre-sudo phase are not artificially constructed by the user in such a way that would allow them to execute more than we want them to. Maybe the `sudo` rule could restrict the first few arguments to be `podman run ...` instead of `*`.... No matter which approach I consider, I find very obvious holes...

# Background

df8885777d2257a6271ca1a6461673d1d96e99d6

# Issue

Running the script requires the user to be a full sudoer, which is not a great requirement to have.

# Solution

Create a new group, `ilab`, and add a sudoers rule that allows users in the `ilab` group to run the wrapper-script as sudo without a password prompt.

Users will not actually run the script as sudo directly, but the script will re-run itself as sudo.

### DIFF
--- a/training/ilab-wrapper/ilab
+++ b/training/ilab-wrapper/ilab
@@ -1,6 +1,21 @@
 #!/bin/bash
 
-echo-err() { echo "$@" >&2; }
+# This script is a wrapper around the ilab command in a podman container that
+# tries to make the experience as seamless as possible. Currently accessing the
+# pre-pulled ilab image requires root access, so the script will attempt to
+# re-run itself with sudo. A RHEL AI system should come preloaded with an ilab
+# group that is allowed to run this script as sudo without a password prompt,
+# so make sure that the user running this script is in the ilab group to avoid
+# the prompt. Note that it's best that you not run this script yourself as sudo,
+# it's best you let the script to re-run itself as sudo.
+#
+# For security reasons we map root UID 0 inside the container to the current
+# user's UID (and all the other subuids to the user's /etc/subuid range) so
+# that we're effectively running the container as the current user.
+#
+# In the future, we will run podman as the current user, once we figure a
+# reasonable way for the current user to access the root's user container
+# storage.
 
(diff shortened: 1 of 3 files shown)


training/ilab-wrapper/ilab
```diff
@@ -18,96 +33,115 @@ function verify_range() {
...
+        #
+        # Also we must conditionally set the environment variable, because if
+        # the variable is not set, we don't want to pass an empty value to the
+        # container, this could lead to unexpected behavior from instructlab.
+        #
+        # Also we don't want to simply use sudo --preserve-env=VARNAME,VARNAME,...
+        # because this requires special sudo permissions that are not present
+        # for all users by default.
+        if [[ -n "${!ENV_VAR}" ]]; then
+            PODMAN_ENV_FLAGS+=("--env" "$ENV_VAR=${!ENV_VAR}")
```
n1hility, line 120: We can't use the = syntax (see recent fix for explanation https://github.com/containers/ai-lab-recipes/pull/731)

This is no longer needed since the --preserve-env  line above (added in above fix) carries the env from the calling environment into sudo, and then the podman exec will reuse the correct value from there.

omertuc: See comment:

```bash
        # Also we must conditionally set the environment variable, because if
        # the variable is not set, we don't want to pass an empty value to the
        # container, this could lead to unexpected behavior from instructlab.
        #
        # Also we don't want to simply use sudo --preserve-env=VARNAME,VARNAME,...
        # because this requires special sudo permissions that are not present
        # for all users by default.
```

omertuc:
Decided to look into it in the future, closing for now

PR #744, opened 2024-08-08
fix: fix image name on chatbot ai-lab.yaml

lstocchi:
It fixes the image name on the chatbot ai-lab.yaml. It should be `llamacpp_python:latest`


rhatdan (approved):
LGTM

PR #746, opened 2024-08-08
feat(nvidia-bootc): Retag instructlab image if INSTRUCTLAB_IMAGE_RETAG is specified

pastequo:
When instructlab is pulled by sha, then no tag is displayed.

If the build-arg INSTRUCTLAB_IMAGE_RETAG is specified, the image will be retag after being pulled


lmilbaum (approved):
LGTM

PR #747, opened 2024-08-08
fix(whispercpp): adding missing ownership to user 1001 in app directory

### LINKED ISSUE #743: whisper image is broken on main
When testing https://github.com/containers/podman-desktop-extension-ai-lab/pull/1512 I got the following error when running the curl command:

## ghcr.io/containers/whispercpp:latest

```bash
$: curl -X POST "http://localhost:38285/inference" -F "file=@./004NTWY_U1_RC.mp3"
{"error":"FFmpeg conversion failed."}
```

## whisper image from `v1.1.3`

```bash
$: curl -X POST "http://localhost:38285/inference" -F "file=@./004NTWY_U1_RC.mp3"
{"text":" Unit 1, page 14. Real conversations.\n Hello. Hi. What's your name?\n Claudia, what's your name?
```

axel7083:
Fixes https://github.com/containers/ai-lab-recipes/issues/743

The user 1001 needs permission to write to the /app directory as it creates a temp file in the current directory.

See https://github.com/ggerganov/whisper.cpp/blob/fe36c909715e6751277ddb020e7892c7670b61d4/examples/server/server.cpp#L680 for ref

### DIFF
--- a/model_servers/whispercpp/base/Containerfile
+++ b/model_servers/whispercpp/base/Containerfile
@@ -19,7 +19,8 @@ COPY --from=builder /app /app
 COPY --from=mwader/static-ffmpeg:6.1.1 /ffmpeg /bin/
 COPY --from=mwader/static-ffmpeg:6.1.1 /ffprobe /bin/
 
-COPY --chown=1001:0 --chmod=755 src /app
+COPY --chown=0:0 --chmod=755 src /app
+RUN chown 1001:1001 /app
 
 USER 1001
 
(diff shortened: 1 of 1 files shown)


axel7083:
cc @johnmcollier

rhatdan:
Just setting the directory to 1001 does not modify the ownership of any content in the directly, which would be owned by root?

Do you want the entire directory to be owed by UID=1001?

Gregory-Pereira (approved):
good catch, /lgtm

axel7083:
> Just setting the directory to 1001 does not modify the ownership of any content in the directly, which would be owned by root?

Yes that's not an issue, as user 1001 does not need to change anything existing, it just need to be able to write some file in it.

> Do you want the entire directory to be owed by UID=1001?

I want the USER 1001 to be able to write a file in the directory, precisely the `whisper_server_temp_file.wav` they uses for tmp file conversion.

PR #750, opened 2024-08-09
Configure model for the rag recipe

maysunfaisal:
Make the same changes like https://github.com/containers/ai-lab-recipes/pull/519 for the rag recipe to allow model name, this lets us use the vLLM model servers on OpenShift with GPUs

Before:
Screenshot 2024-08-08 at 6 44 29 PM

After:
Screenshot 2024-08-08 at 7 08 06 PM


rhatdan (approved):
LGTM

PR #755, opened 2024-08-12
Move --target-arch to image builder args

javipolo:
Move --target-arch to be an argument to `bootc-image-builder` instead of as an argument to `podman`

### DIFF
--- a/training/common/Makefile.common
+++ b/training/common/Makefile.common
@@ -89,7 +89,6 @@ check-umask:
 bootc-image-builder:
 	mkdir -p build/store build/output
 	podman run \
-	  $(ARCH:%=--target-arch %) \
 	  $(AUTH_JSON:%=-v %:/run/containers/0/auth.json) \
 	  --rm \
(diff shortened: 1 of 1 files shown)


rhatdan:
Looks like you have an ancient version of bootc-image-builder?

podman run quay.io/centos-bootc/bootc-image-builder:latest --help | grep arch
      --target-arch string    build for the given target architecture (experimental)

javipolo:
> Looks like you have an ancient version of bootc-image-builder?
> 
> podman run quay.io/centos-bootc/bootc-image-builder:latest --help | grep arch --target-arch string build for the given target architecture (experimental)

I just saw the problem now :facepalm: 
we were passing `--target-arch` as an argument to `podman` instead of passing it as an argument to the `bootc-image-builder` container :facepalm: :facepalm: 

I'm going to change the PR description to reflect that :)

rhatdan (approved):
LGTM

PR #757, opened 2024-08-13
Remove retagging from ilab image

enriquebelarte:
Retag is not needed as we are using static tags now.


lmilbaum (approved):
LGTM

lmilbaum:
@pastequo Please review the changes before it is merged

rhatdan:
Are you sure the image in the additional store ends up labeled correctly?

pastequo:
The retag was instaured because without it "podman images" or equivalent was displaying `<none>`

enriquebelarte:
@pastequo  If I'm not mistaken the retag was needed because digest tags made the `podman images ` output show tag field as `<none>` .
We're not using digest tags now but static ones that's why my guess is that the code is not needed anymore.

rhatdan:
LGTM

PR #759, opened 2024-08-14
Modify condition on NVIDIA FabricManager service

fabiendupont:
The `/dev/nvswitchctl` device is created by the NVIDIA Fabric Manager service, so it cannot be a condition for the `nvidia-fabricmanager` service.

Looking at the NVIDIA driver startup script for Kubernetes, the actual check is the presence of `/proc/driver/nvidia-nvswitch/devices` and the fact that it's not empty [1].

This change modifies the condition to
`ConditionDirectoryNotEmpty=/proc/driver/nvidia-nvswitch/devices`, which verifies that a certain path exists and is a non-empty directory.

[1] https://gitlab.com/nvidia/container-images/driver/-/blob/main/rhel9/nvidia-driver?ref_type=heads#L262-269


fabiendupont:
I have tested the new condition on machines with and without an NVLink switch and it works as expected. If there is no NVLink switch, the unit is skipped. If there is an NVLink switch, it is started.

javipolo:
LGTM

Also, tested on a bare metal nvidia node :)

javipolo (approved):

PR #760, opened 2024-08-14
fix: backend property value

### LINKED ISSUE #758: `backend` property in `ai-lab.yaml` file
In the `ai-lab.yaml` files there a `backend` property which is not matching the backend property of the catalog in AI Lab. This field is `llama` for every recipe.

https://github.com/containers/ai-lab-recipes/blob/5db0a4dd83c7a86aa220daba63c7e87eeddba0ae/recipes/natural_language_processing/rag/ai-lab.yaml#L11-L12

In whispercpp it is also llama (weird?)

https://github.com/containers/ai-lab-recipes/blob/8b5adaab7e29353af2d26c885bf032d4593efab1/recipes/audio/audio_to_text/ai-lab.yaml#L11-L12

In AI Lab we have a backend property for the catalog, defining for models and recipe which backend they need, the values allowed are `llama-cpp`, `whisper-cpp` and `none` see [InferenceType](https://github.com/containers/podman-desktop-extension-ai-lab/blob/8563129531a0850365f724c5645abbb3c9e30ecf/packages/shared/src/models/IInference.ts#L20) for details

With the on-going work on AI Lab to make the recipes uses the existing InferenceServer instance, it would be better to unify those values to have them match

axel7083:
## Description

Replace the hardcoded and unused `backend: ['llama']` property to match the value we are using in AI Lab

## Related issue

Fixes https://github.com/containers/ai-lab-recipes/issues/758
Related to https://github.com/containers/podman-desktop-extension-ai-lab/issues/1542

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,7 +43,8 @@ application:
       contextdir: ../../../model_servers/llamacpp_python
       containerfile: ./base/Containerfile
       model-service: true
-      backend: llama-cpp
+      backend: 
+        - llama-cpp
       arch:
         - arm64
         - amd64
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,8 +43,7 @@ application:
       contextdir: ../../../model_servers/llamacpp_python
       containerfile: ./base/Containerfile
       model-service: true
-      backend: 
-        - llama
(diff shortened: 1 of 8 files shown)


rhatdan (approved):
LGTM

PR #766, opened 2024-08-22
Share the additional container storage

fabiendupont:
We need to share container image storage between rootless users, so that
we don't need `sudo` and we don't duplicate the `instructlab` image.
This change follows the Red Hat solution to
[create additional image store for rootless users](https://access.redhat.com/solutions/6206192).

The `/usr/lib/containers/storage` folder can be read by anyone and new
users will inherit a default configuration via `/etc/skel` that
configures the additional storage.

The `ilab` wrapper is also modified to remove the impersonation code and
not use `sudo` anymore.

### DIFF
--- a/training/nvidia-bootc/Containerfile
+++ b/training/nvidia-bootc/Containerfile
@@ -187,12 +187,15 @@ VOLUME /var/lib/containers
 RUN --mount=type=secret,id=${INSTRUCTLAB_IMAGE_PULL_SECRET}/.dockerconfigjson \
     if [ -f "/run/.input/instructlab-nvidia/oci-layout" ]; then \
          IID=$(podman --root /usr/lib/containers/storage pull oci:/run/.input/instructlab-nvidia) && \
-         podman --root /usr/lib/containers/storage image tag ${IID} ${INSTRUCTLAB_IMAGE}; \
+         podman --root /usr/lib/containers/storage --storage-opt overlay.force_mask=shared image tag ${IID} ${INSTRUCTLAB_IMAGE}; \
     elif [ -f "/run/secrets/${INSTRUCTLAB_IMAGE_PULL_SECRET}/.dockerconfigjson" ]; then \
-         IID=$(sudo podman --root /usr/lib/containers/storage pull --authfile /run/secrets/${INSTRUCTLAB_IMAGE_PULL_SECRET}/.dockerconfigjson ${INSTRUCTLAB_IMAGE}); \
+         IID=$(sudo podman --root /usr/lib/containers/storage pull --storage-opt overlay.force_mask=shared --authfile /run/secrets/${INSTRUCTLAB_IMAGE_PULL_SECRET}/.dockerconfigjson ${INSTRUCTLAB_IMAGE}); \
     else \
-         IID=$(sudo podman --root /usr/lib/containers/storage pull ${INSTRUCTLAB_IMAGE}); \
-    fi
+         IID=$(sudo podman --root /usr/lib/containers/storage --storage-opt overlay.force_mask=shared pull ${INSTRUCTLAB_IMAGE}); \
+    fi && \
+    chmod -R a+rX /usr/lib/containers
+
+COPY containers-storage.conf /etc/skel/.config/containers/storage.conf
 
 RUN podman system reset --force 2>/dev/null
 
--- a/training/nvidia-bootc/containers-storage.conf
+++ b/training/nvidia-bootc/containers-storage.conf
@@ -1,0 +1,17 @@
+[storage]
+driver = "overlay"
+
+[storage.options]
+size = ""
+remap-uids = ""
+remap-gids = ""
+ignore_chown_errors = ""
+remap-user = ""
+remap-group = ""
+skip_mount_home = ""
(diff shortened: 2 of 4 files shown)


training/nvidia-bootc/containers-storage.conf
```diff
@@ -0,0 +1,13 @@
...
+    size = ""
+    remap-uids = ""
+    remap-gids = ""
+    ignore_chown_errors = ""
+    remap-user = ""
+    remap-group = ""
+    skip_mount_home = ""
+    mount_program = "/usr/bin/fuse-overlayfs"
+    mountopt = ""
+additionalimagestores = [ "/usr/lib/containers/storage",]
```
eranco74, line 14: We will need to allow non-root users access (at lest read) to this path.
Probably need to sort out selinux as well.

eranco74 (commented):
I really like this approach, just need to sort out the permissions.
If we allow only read access we should add an additional storage path where users may write to.

training/nvidia-bootc/containers-storage.conf
fabiendupont: The read access is set in the Containerfile. We could limit to a specific group, too.

training/nvidia-bootc/Containerfile
```diff
@@ -191,6 +191,9 @@ RUN --mount=type=secret,id=${INSTRUCTLAB_IMAGE_PULL_SECRET}/.dockerconfigjson \
     else \
          IID=$(sudo podman --root /usr/lib/containers/storage pull ${INSTRUCTLAB_IMAGE}); \
     fi
+    chmod -R a+rx /usr/lib/containers/storage
```
omertuc: This was our original approach but we abandoned it because this makes all the files even inside the container executable

omertuc: https://github.com/containers/ai-lab-recipes/pull/705

fabiendupont: We could refine with `find` to put the `x` only on the folders. Would that be better?

omertuc: There was [this](https://github.com/containers/ai-lab-recipes/pull/705#discussion_r1698613551) suggestion but I never tried it, IDK

fabiendupont: @cgwalters's suggestion to use `chmod a+rX` is nice. It will avoid adding `x` permission to config files or example scripts.

rhatdan:
@giuseppe PTAL

training/nvidia-bootc/Containerfile
```diff
@@ -190,7 +190,10 @@ RUN --mount=type=secret,id=${INSTRUCTLAB_IMAGE_PULL_SECRET}/.dockerconfigjson \
          IID=$(sudo podman --root /usr/lib/containers/storage pull --authfile /run/secrets/${INSTRUCTLAB_IMAGE_PULL_SECRET}/.dockerconfigjson ${INSTRUCTLAB_IMAGE}); \
     else \
          IID=$(sudo podman --root /usr/lib/containers/storage pull ${INSTRUCTLAB_IMAGE}); \
-    fi
+    fi && \
+    chmod -R a+rX /usr/lib/containers
```
giuseppe, line 196: are you also setting `force_mask=shared` in the `storage.conf` file?

If you do so, then podman will store the original permissions in an extended attribute, that it is used by fuse-overlays

fabiendupont: Currently, it is not set. Should it be in the global `storage.conf` or it could be in the user config?

giuseppe: sorry for the delay.

It must be in the configuration file for the user that is creating the storage (i.e. runs `podman pull`).  Also, when using `--root`, it tells Podman to ignore any other setting from the `storage.conf`, so in this case, it must be provided on the command line as a `--storage-opt overlay.force_mask=shared`

fabiendupont: Thanks @giuseppe. I updated the user `storage.conf` template and the Containerfile. Would you mind reviewing again, please?

rhatdan:
@giuseppe PTANL

training/nvidia-bootc/Containerfile
giuseppe: Thanks, LGTM

giuseppe (approved):
LGTM

rhatdan:
@fabiendupont still something we want to merge?  Will this work well with rootful containers as well, without using fuse-overlay?

fabiendupont:
@rhatdan, yes, that's still something we want to merge. It removes the requirement to use `sudo`, so it's a step in the right direction.
IIUC, the rootful containers shouldn't be impacted. The overlay config only applies to standard users and the `instructlab` container that is loaded during the build. If someone adds container images via layering, the won't be readable by users, unless the `chmod -R a+X` command is run again.

rhatdan:
LGTM

training/nvidia-bootc/Containerfile
omertuc: Note that bootc does not preserve these extended arguments, you will lose them

PR #767, opened 2024-08-26
updating AMD  bootc image

yevgeny-shnaidman:
1) using OOT driver and firmware instead of in-tree 
2) moving to DKMS and ROCM 6.1.2 version


fabiendupont:
Instead of commenting here, I submitted a pull request in @yevgeny-shnaidman's repository: https://github.com/yevgeny-shnaidman/ai-lab-recipes/pull/1.

fabiendupont:
LGTM

training/amd-bootc/Containerfile
```diff
@@ -1,43 +1,49 @@
-# Define the images to be used
 ARG INSTRUCTLAB_IMAGE="quay.io/ai-lab/instructlab-amd:latest"
 ARG BASEIMAGE="quay.io/centos-bootc/centos-bootc:stream9"
+ARG DRIVER_TOOLKIT_IMAGE="quay.io/ai-lab/nvidia-builder:latest"
```
rhatdan, line 3: Does this make sense?  Using nvidia-builder on amd-bootc?

yevgeny-shnaidman: this is a driver-toolkit image, which contains all the package needed for kernel module compilation. So, it is related to a kernel version, and does not contain any packages that might be specific to nvidia. The name unfortunate, but this is what we have right now

fabiendupont: We have discussed moving the bootc images out of the ai-lab-recipes repository. One of the changes would be to have a `quay.io/<org>/driver-toolkit:<kernel-version>`. The driver toolkit images could be hosted in the quay.io/fedora and quay.io/centos orgs.

PR #770, opened 2024-08-28
Preserve PYTORCH_CUDA_ALLOC_CONF to ilab container

javipolo:


javipolo:
Setting on hold until we agree if we want users to be tweaking this kind of things

PR #782, opened 2024-09-12
Bootc - Add git and git-lfs packages

fabiendupont:
When working with AI/ML recipes, it is frequent to pull versioned software and data from Git repositories. This change adds the `git` and `git-lfs` packages.

### DIFF
--- a/training/amd-bootc/Containerfile
+++ b/training/amd-bootc/Containerfile
@@ -36,6 +36,8 @@ RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-AMD-ROCM \
     && dnf install -y \
         amd-smi \
         cloud-init \
+        git \
+        git-lfs \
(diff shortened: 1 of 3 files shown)


training/amd-bootc/cmd
```diff
@@ -0,0 +1 @@
+make amd FROM=registry.redhat.io/rhel9/rhel-bootc:9.4 DRIVER_TOOLKIT_IMAGE=registry.redhat.io/rhel9/driver-toolkit:5.14.0-427.31.1.el9_4
```
rhatdan: I thought we were trying to avoid hard coding a rhel image into this, so that this upstream project can be used by anyone?

fabiendupont: That's a leftover from tests. I removed it.

rhatdan (approved):
LGTM

PR #784, opened 2024-09-16
Update environment preservation for Podman with python https ca control variables

relyt0925:
Add SSL_CERT_FILE and SSL_CERT_DIR to the preserved environment variables and ensure they are passed to Podman. This change ensures that SSL certificates are correctly handled within the container environment. 

Provides path for sdg to interact with teacher models with certificates issued from private CAs (https://github.com/instructlab/sdg/pull/106).


rhatdan:
LGTM

PR #785, opened 2024-09-17
driver-toolkit make target fixes.

eelgaev:
Added `BASEIMAGE` argument to driver-toolkit make target and fixed a target name typo.


rhatdan:
LGTM

PR #791, opened 2024-09-23
Some fixes to intel Containerfile

enriquebelarte:
- Removed some packages from deletion as not all devel packages can be removed as they're dependencies.
- Added workaround for libdnf issue when running dnf in Konflux.
- Added some bits needed for instructlab config.


rhatdan:
LGTM

PR #793, opened 2024-09-24
Fix intel bootc script variable

enriquebelarte:
Fix secret variable declaration at script.


rhatdan:
I think we have an outstanding issue in buildah to support environment variable secrets.

LGTM

PR #795, opened 2024-09-27
amd-bootc: Share the additional container storage

fabiendupont:
We need to share container image storage between rootless users, so that we don't need `sudo` and we don't duplicate the `instructlab` image. This change follows the Red Hat solution to
[create additional image store for rootless
users](https://access.redhat.com/solutions/6206192).

The `/usr/lib/containers/storage` folder can be read by anyone and new users will inherit a default configuration via `/etc/skel` that configures the additional storage.

The `ilab` wrapper is also modified to remove the impersonation code and not use `sudo` anymore.

Follow-up on #766


fabiendupont:
@rhatdan, @giuseppe

giuseppe (approved):
LGTM

rhatdan:
LGTM

PR #798, opened 2024-10-01
AMD bootc - amdgpu kernel modules build process

eelgaev:
This PR implements a custom build process for out-of-tree amdgpu modules, instead of using dkms.

- The same version of the out-of-tree source code is downloaded, and a script is executed to build the modules.
- After the kernel modules are built, they are packaged into an RPM and installed in the final amd-bootc image as a package.


rhatdan:
@Okoyl what is going on with this PR?

PR #800, opened 2024-10-08
feat: add JavaScript/Node.js based chatbot recipe

mhdawson:
I don't think we'd want to duplicate all of the examples but having a Node.js example for the chatbot and for the RAG recipe is what I'd think would be a good place to be.  This PR is the first step of adding the chatbot example in Node.js

It might need a few tweaks/testing of the framework stuff to build/package under quadlet but I have updated most things I think are needed and done what testing I could do without a bit more guidance.


mhdawson:
@Gregory-Pereira as discussed FYI.

mhdawson:
@Gregory-Pereira wondering what the next step might be on moving this forward?

recipes/natural_language_processing/chatbot-nodejs/ai-lab.yaml
```diff
@@ -0,0 +1,27 @@
...
+  type: language
+  name: ChatBot_nodejs
+  description: Chat with a model service in a web frontend.
+  containers:
+    - name: llamacpp-server
+      contextdir: ../../../model_servers/llamacpp_python
+      containerfile: ./base/Containerfile
+      model-service: true
+      backend:
+        - llama
```
Gregory-Pereira: Shouldnt this be:
```suggestion
        - llama-cpp
```
as per: https://github.com/containers/ai-lab-recipes/blob/main/recipes/natural_language_processing/chatbot/ai-lab.yaml#L12

recipes/natural_language_processing/chatbot-nodejs/provision/playbook.yml
Gregory-Pereira, line 1: You dont actually need this directory - it is for the E2E bootc framework which this will not need to make it into. You can remove this.

recipes/natural_language_processing/chatbot-nodejs/provision/requirements.yml
Gregory-Pereira, line 1: Same as above, remove this file + directory.

Gregory-Pereira (commented):
Generally looks pretty close, just some small fixes. I will also want to test this locally before I approve for merging, unless you have done this already

recipes/natural_language_processing/chatbot-nodejs/provision/requirements.yml
mhdawson: @Gregory-Pereira removed the directory and the 2 files

recipes/natural_language_processing/chatbot-nodejs/ai-lab.yaml
mhdawson: Looks like it was changed since I put the PR together, accepted the suggestion.

mhdawson:
@Gregory-Pereira I had tested it out to the extend that I could, but I think you testing locally would be good as a sanity check.  It was also quite a while ago that I ran that testing./

rhatdan:
@Gregory-Pereira this is waiting on you?

Gregory-Pereira (approved):
/LGTM

PR #815, opened 2024-11-20
fix: support application/octet-stream file type

### LINKED ISSUE #814: summarizer app not accepting markdown file
When I upload an MD file to the Summarizer app, I get an error. After looking at the source and debugging, I can see that the detected `file_type` for the markdown file is `application/octet-stream` and not `text/markdown`

Image

feloy:
Fixes #814


rhatdan:
LGTM

PR #821, opened 2024-11-22
fix: update chatbot-nodejs to bind to localhost

### LINKED ISSUE #819: chatbot-nodejs does not listen on localhost
Podman AI Lab uses a healthcheck using localhost to detect container health

jeffmaury:
Fixes #819

### DIFF
--- a/recipes/natural_language_processing/chatbot-nodejs/app/Containerfile
+++ b/recipes/natural_language_processing/chatbot-nodejs/app/Containerfile
@@ -53,4 +53,4 @@ EXPOSE 8051
 
 ENV PORT 8051
 
-CMD ["node", "server"]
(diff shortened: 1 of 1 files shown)


rhatdan (approved):
LGTM

mhdawson:
@jeffmaury looks like you were working to get the nodejs-chatbot example working in podman-ai-lab at the same time as I was along with @Gregory-Pereira.

I've fixed up some ports, and I have a running version if I revert this change to bind to localhost. I think either it will need to bind to all ports not just local host (as otherwise the app was not available externally) or revert back to what was there before.

I think the issue related to the health checks might be that I might not have included them, but not sure as it was quite a while since I first wrote the code.

mhdawson:
@jeffmaury do you mind if I revert your change and then take a look at why the health checks are not reporting healthy ?

PR #822, opened 2024-11-22
fix: fix the port used by Node.js chatbot example

mhdawson:


mhdawson:
Fix wrong port number used in the Node.js chatbot example. Many thanks to @Gregory-Pereira for helping me find this.

mhdawson:
Also includes a fix for another typo that was not affecting the deployment in ai lab but was different from the python version.

Gregory-Pereira (approved):
/lgtm and tested

PR #825, opened 2024-11-25
fix: bind to both local and externally visible IPs

mhdawson:
- allows both health check and app to work correctly


mhdawson:
@jeffmaury this tweak to you update allows both the app and the health check to work.

mhdawson:
@Gregory-Pereira one last PR and I think all should be good

jeffmaury:
LGTM

Gregory-Pereira (approved):

PR #828, opened 2024-12-03
Documentation updates

sallyom:
A few cleanup items and updating RAG README for running with milvus.

@rhatdan @MichaelClifford

### DIFF
--- a/model_servers/llamacpp_python/README.md
+++ b/model_servers/llamacpp_python/README.md
@@ -19,8 +19,9 @@ The [base image](../llamacpp_python/base/Containerfile) is the standard image th
 To build the base model service image:
 
 ```bash
-make -f Makefile build
+make build
 ```
+
 To pull the base model service image:
 
 ```bash
@@ -33,8 +34,9 @@ podman pull quay.io/ai-lab/llamacpp_python
 The [Cuda image](../llamacpp_python/cuda/Containerfile) include all the extra drivers necessary to run our model server with Nvidia GPUs. This will significant speed up the models response time over CPU only deployments.
 
 To Build the the Cuda variant image:
+
 ```bash
-make -f Makefile build-cuda
+make build-cuda
 ```
 
 To pull the base model service image:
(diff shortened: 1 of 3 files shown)


sallyom:
~this PR disables vulkan image builds: https://github.com/containers/ai-lab-recipes/issues/829~
removed that commit

rhatdan:
Do the Vulkan builds still need to be disabled?

rhatdan:
Please remove and update your repo, vulkan was just updated.

sallyom:
> Do the Vulkan builds still need to be disabled?

nope, removed it - I was hoping someone else would fix the vulkan build and they did :)

model_servers/llamacpp_python/README.md
```diff
@@ -88,7 +90,8 @@ Place all models in the [models](../../models/) directory.
 You can use this snippet below to download the default model:
 
 ```bash
-make -f Makefile download-model-granite
+cd ../../models
+make download-model-granite
```
Gregory-Pereira: Both of these are technically correct, what you have and what I had previously. This is because I have an [override command in the makefile](https://github.com/containers/ai-lab-recipes/blob/main/model_servers/llamacpp_python/Makefile#L28-L31) which will call that make command in models. You can change this if you want but I think it is more correct to use the override.

Gregory-Pereira (commented):
Generally /LGTM. Just once nitpicky comment but everything looks fine. That being said I havent tested any of this

model_servers/llamacpp_python/README.md
sallyom: I did not notice that in the llamacpp_python/Makefile, updated!

jeffmaury (changes requested):
Most of the changes in this PR are unrelated to the purpose explained in the title so it should be splitted into 2 different PRs for better readability

sallyom:
@jeffmaury ptal - this PR is now only documentation updates. The rag/app/manage_vectordb.py import update I'll open in  https://github.com/containers/ai-lab-recipes/pull/838.

jeffmaury (approved):
LGTM

Issue #811, opened 2024-11-11
ai-lab/granite-7b-lab not available on quay.io


grafuls:
When running the recipe for the ChatBot the following command fails with:
```
$ podman kube play build/chatbot.yaml
Trying to pull quay.io/ai-lab/granite-7b-lab:latest...
Error: initializing source docker://quay.io/ai-lab/granite-7b-lab:latest: reading manifest latest in quay.io/ai-lab/granite-7b-lab: unauthorized: access to the requested resource is not authorized
```

Possible solution replace with the available `ai-lab/merlinite-7b-lab`

sallyom:
pushed new `quay.io/ai-lab/granite-7b-lab:latest` image, should be available now!

PR #830, opened 2024-12-03
allow update to model images

sallyom:
@Gregory-Pereira I'm not sure why the models images don't get pushed but I think this will allow to manually push them - currently the granite-7b-lab disappeared from quay.

see https://github.com/containers/ai-lab-recipes/issues/811


Gregory-Pereira (approved):
/lgtm

PR #831, opened 2024-12-04
update models build & push workflow

sallyom:
@Gregory-Pereira another fix for the models build workflow, ty!


Gregory-Pereira (approved):
yup we discussed this, /lgtm

PR #834, opened 2024-12-04
feat: add Node.js RAG recipe

mhdawson:


mhdawson:
I've tested this with podman ai lab and all works well there.

I  also test running with `podman kube play build/rag-nodejs.yaml`.  That also works provided I set the chroma db container to ghcr.io/chroma-core/chroma:0.5.16 to match the version that podman ai desktop builds. 
* The chromadb container used based on the default in the shared Makefiles, comes from quay.io is much older (0.4.4) as it looks like the quay container for chromadb has not been updated for 8 months or so.    
* I think the old version of chromadb in the Makefiles is not an issue with this new recipe but something that should be addressed separately. Let me know if I should open an issue on the repo for that or if its a known issue.

mhdawson:
@Gregory-Pereira the second Node.js recipe, it should look familar as it reuses part of the Node.js chatbot recipe.

recipes/natural_language_processing/rag-nodejs/Makefile
```diff
@@ -0,0 +1,67 @@
...
+
+include ../../common/Makefile.common
+
+RECIPE_BINARIES_PATH := $(shell realpath ../../common/bin)
+RELATIVE_MODELS_PATH := ../../../models
+RELATIVE_TESTS_PATH := ../tests
+
+.PHONY: run-chromadb
+run: 
+	podman run -it -p $(CHROMADB_PORT):$(CHROMADB_PORT) -e CHROMADB_ENDPOINT=http://10.88.0.1:8000/v1 ${CHROMADB_IMAGE}
```
Gregory-Pereira, line 14: This stuff should be moved to the [chroma vectordb makefile](https://github.com/containers/ai-lab-recipes/blob/main/vector_dbs/chromadb/Makefile) and called via an override from here.

recipes/natural_language_processing/rag-nodejs/Makefile
```diff
@@ -0,0 +1,67 @@
...
+	  --build-arg SERVER_IMAGE=$(SERVER_IMAGE) \
+	  --build-arg "SSHPUBKEY=$(SSH_PUBKEY)" \
+	  -f bootc/$(CONTAINERFILE) \
+	  -t ${BOOTC_IMAGE} .
+	@echo ""
+	@echo "Successfully built bootc image '${BOOTC_IMAGE}'."
+	@echo "You may now convert the image into a disk image via bootc-image-builder"
+	@echo "or the Podman Desktop Bootc Extension.  For more information, please refer to"
+	@echo "   * https://github.com/osbuild/bootc-image-builder"
+	@echo "   * https://github.com/containers/podman-desktop-extension-bootc"
```
Gregory-Pereira, line 38: I would rather see this update the [common makefile to allow an extra bootc buildarg](https://github.com/containers/ai-lab-recipes/blob/main/recipes/common/Makefile.common#L85-L105), which would reduce code duplicaiton. What im imagining is some variable like `EXTRA_BOOTC_BUILD_ARGS`:
Makefile.common:
```Makefile.common
EXTRA_BOOTC_BUILD_ARGS ?=
...
.PHONY: bootc
bootc: quadlet growfs
	podman build \
	  $(ARCH:%=--arch %) \
	  $(BUILD_ARG_FILE:%=--build-arg-file=%) \
	  $(FROM:%=--from %) \
	  $(AUTH_JSON:%=-v %:/run/containers/0/auth.json) \
	  --security-opt label=disable \
	  --cap-add SYS_ADMIN \
	  --build-arg MODEL_IMAGE=$(MODEL_IMAGE) \
	  --build-arg APP_IMAGE=$(APP_IMAGE) \
	  --build-arg SERVER_IMAGE=$(SERVER_IMAGE) \
	  --build-arg "SSHPUBKEY=$(SSH_PUBKEY)" \
      ${EXTRA_BOOTC_BUILD_ARGS} \
	  -f bootc/$(CONTAINERFILE) \
	  -t ${BOOTC_IMAGE} bootc
	@echo ""
	@echo "Successfully built bootc image '${BOOTC_IMAGE}'."
	@echo "You may now convert the image into a disk image via bootc-image-builder"
	@echo "or the Podman Desktop Bootc Extension.  For more information, please refer to"
	@echo "   * https://github.com/osbuild/bootc-image-builder"
	@echo "   * https://github.com/containers/podman-desktop-extension-bootc"
	
```

Then from the recipe you would call it like: `EXTRA_BOOTC_BUILD_ARGS="--build-arg something=something_else" make bootc`, etc.

recipes/natural_language_processing/rag-nodejs/Makefile
```diff
@@ -0,0 +1,67 @@
...
+	    -e "s|APP|${APP}|g" \
+	    quadlet/${APP}.image \
+	    > build/${APP}.image
+	sed -e "s|SERVER_IMAGE|${SERVER_IMAGE}|" \
+	    -e "s|APP_IMAGE|${APP_IMAGE}|g" \
+	    -e "s|MODEL_IMAGE|${MODEL_IMAGE}|g" \
+	    -e "s|CHROMADB_IMAGE|${CHROMADB_IMAGE}|g" \
+	    quadlet/${APP}.yaml \
+	    > build/${APP}.yaml
+	cp quadlet/${APP}.kube build/${APP}.kube
```
Gregory-Pereira, line 59: Same as above, lets find a non-invasive way to extend the common option to suit your needs here. This is an important pattern because it helps reduce complexity when debugging, leaving 1 section of code to traverse and step through rather than 2.

recipes/natural_language_processing/rag-nodejs/Makefile
```diff
@@ -0,0 +1,67 @@
...
+
+# rag requires custom bootc-run because it uses an extra port for chromadb
+# (other apps use ../../common/Makefile.common target)
+.PHONY: bootc-run
+bootc-run:
+	podman run -d --rm --name $(APP)-bootc -p 8080:8501 -p 8090:8000 --privileged \
+	  $(AUTH_JSON:%=-v %:/run/containers/0/auth.json) \
+	  $(BOOTC_IMAGE) /sbin/init
```
Gregory-Pereira, line 67: Same as above

recipes/natural_language_processing/rag-nodejs/bootc/Containerfile
```diff
@@ -0,0 +1,48 @@
...
+# public key into the image, allowing root access via ssh.
+RUN set -eu; mkdir -p /usr/ssh && \
+    echo 'AuthorizedKeysFile /usr/ssh/%u.keys .ssh/authorized_keys .ssh/authorized_keys2' >> /etc/ssh/sshd_config.d/30-auth-system.conf && \
+    echo ${SSHPUBKEY} > /usr/ssh/root.keys && chmod 0600 /usr/ssh/root.keys
+
+ARG RECIPE=rag-nodejs
+ARG MODEL_IMAGE=quay.io/ai-lab/granite-7b-lab:latest
+ARG APP_IMAGE=quay.io/ai-lab/${RECIPE}:latest
+ARG SERVER_IMAGE=quay.io/ai-lab/llamacpp_python:latest
+ARG CHROMADBImage=quay.io/ai-lab/chromadb
```
Gregory-Pereira: I dont believe this is correct, should be: 
```suggestion
ARG CHROMADB_IMAGE=quay.io/ai-lab/chromadb:latest
```

Gregory-Pereira (commented):
Overall great PR. I just had some process points. If there is a reason to rush this in like a blog before EOY or some other time dependent deliverable we could postpone those to a follow up PR, but if not Id prefer we implement through common makefile.

mhdawson:
@Gregory-Pereira in terms of the suggested changes, those are all to files which are almost a direct copy of those being used for the existing rag recipe. Is that something that should be part of this PR or a cleanup that covers both?   I have enough understanding to copy/tweak those files, but would a bigger lift to understand well enough to refactor.

I just copied them over and updated to use `rag-nodesjs` instead of `rag` for the most part.

There is no major rush, but would like to avoid having other PRs having me to rework.

recipes/natural_language_processing/rag-nodejs/bootc/Containerfile
mhdawson: Its a direct copy from the existing rag recipe - https://github.com/containers/ai-lab-recipes/blob/7fbdf03a8dadbf5a41da283fdc1c07ea8d1ec61a/recipes/natural_language_processing/rag/bootc/Containerfile#L21

mhdawson:
I'm off on PTO until Dec 23 so if I don't respond to comments/questions that's why. I'll check in when I return on the 23rd.

mhdawson:
I'm back for the next few days but then out again until Jan 2nd. In any case I expect lots of people will be out and will check in again when I'm back in Jan.

recipes/natural_language_processing/rag-nodejs/Makefile
Gregory-Pereira: Seems like this is copied from the other rag example, however we should have extended the base in other cases. Since I dont want to set  a new precendent, we will consider these a non issue for now and refactor.

mhdawson:
@Gregory-Pereira I pushed a commit with the changes you requested.

I then pushed another commit which set the chromadb container used by quadlet and bootc to match what is used when the recipe is used with the podman ai lab extension. It seems that chromadb is still changing quickly and the client version seems to need to match the chromadb  version in use.  This addresses the issue mentioned eariler and I think is better than changing the common chromadb version that is used for quadlet in the python recipe as it might have an older client  and that might explain why an older container is being use.

After these changes the I test that the recipe still works with the podman desktop ai extension as well as when you follow the instructions to run it with quadlet.

recipes/natural_language_processing/rag-nodejs/bootc/Containerfile
mhdawson: Updated

Gregory-Pereira (approved):
/lgtm

mhdawson:
@Gregory-Pereira many thanks for your help.

PR #839, opened 2025-01-07
fix variable naming conventions

Gregory-Pereira:
This was found during #834. Naming and tagging conventions should be consistent


cooktheryan (approved):

rhatdan (approved):
LGTM

PR #841, opened 2025-01-09
fix: update Rag Node.js recipie to make compatible with later chromadb version

mhdawson:
https://github.com/containers/ai-lab-recipes/pull/806 updated the version of chromadb used with the rag recipe when run with podman ai lab.

Update the versions of Langchain and Chromadb clients to be compatible


mhdawson:
@Gregory-Pereira some some tweaks needed I noticed where needed after testing in podman ai lab extension with updates that had happened in the ai-lab-recipes since I had last rebased what was in the PR to add the node.js RAG recipe.

Gregory-Pereira (approved):
/lgtm

Issue #756, opened 2024-08-12
WIP: Add GraphRAG recipe


srampal:

srampal:
Working on this ...

srampal:
Will be added via this PR which is a WIP
https://github.com/containers/ai-lab-recipes/pull/801

jeffmaury:
Running the recipe in Podman AI Lab leads to an error:

![Image](https://github.com/user-attachments/assets/d11729b1-22aa-47e4-99f2-48ec71558bde)

bmahabirbu:
Fixed current PR

https://github.com/containers/ai-lab-recipes/pull/855

bmahabirbu:
Had to do with permission errors for the work directory similar to the pydantic recipe

PR #853, opened 2025-02-26
Added GraphRag Recipe using Lightrag repo

bmahabirbu:
This recipe uses some advanced features so GPU is recommended to get good response times. Building the image and testing can take a super long time which is unfortunate and has to do with torch. 

Every framework that I've tried with GraphRag implementation is slow so its just the implementation of the actual feature that's the problem. Hopefully in the future this can change. 

Also this recipe is best suited for models with function/tool calling. While models without can work you have to select the "naive" button to get a response back.

Token size limit was also another issue maybe in the future we can have the model service have a larger token size to avoid too many truncations!

What issues does this PR fix or reference?
[fixes https://github.com/containers/ai-lab-recipes/issues/756]

How to test this PR?
Manual test:

cd to graph-rag
make build
make quadlet
podman kube play build/chatbot-pydantic-ai.yal
check localhost::8501

You can also use ramalama backend
make build
make run
ramalama serve your_model -p 8001


rhatdan:
LGTM

srampal:
@bmahabirbu nice work using lightrag to address the graphrag requirement. I'd like to discuss the pros and cons you saw with lightrag vs other options for graphrag. DM me on twitter @sr2357 to discuss details ?

PR #855, opened 2025-02-27
Fixed containerfile build issue

### LINKED ISSUE #756: WIP: Add GraphRAG recipe

bmahabirbu:
Had to edit permissions on workdir 

Fixes #756

### DIFF
--- a/recipes/natural_language_processing/graph-rag/app/Containerfile
+++ b/recipes/natural_language_processing/graph-rag/app/Containerfile
@@ -1,7 +1,11 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-77.1726664316
 
+# Set the working directory
 WORKDIR /graph-rag
-RUN chmod -R 777 /graph-rag
+# Ensure the default user has the correct permissions on the working directory
+RUN chown -R 1001:0 /graph-rag
+
+# Copy the application files
 COPY requirements.txt .  
 COPY rag_app.py .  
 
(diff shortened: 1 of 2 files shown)


rhatdan:
What user is attempting to write to the WORKDIR?

bmahabirbu:
Not a user but the code is writing to the workdir to save the vectordb locally

rhatdan:
Understood, but the code within the container is either running as root, which would be allowed to write to the directory or it is writing as non-root and the ownership of the directory is wrong and needs to be chowned to the right user.

bmahabirbu:
I wonder what the best solution would be here

image

This is the error that was happening without modifying the workdir permissions.

In ai lab recipes we are using podman build --squash-all which fixes this permission issue but building without this argument (like they do in ai lab extensions) it runs in to this problem. I believe squash-all combines all layers and that solves the issue (but im not entirely sure why)

bmahabirbu:
I believe it is writing as non-root. I'm wondering then what the right user should be to chowend it properly

recipes/natural_language_processing/graph-rag/app/Containerfile
```diff
@@ -1,6 +1,7 @@
 FROM registry.access.redhat.com/ubi9/python-311:1-77.1726664316
 
 WORKDIR /graph-rag
+RUN chmod -R 777 /graph-rag
```
jeffmaury: As this is a kind of security hole, I would prefer to have replaced by:

`RUN chmod -R default /graph-rap`

Should be run after the COPY instructions as COPY instructions are running as root and they are the reason why `/grap-rag` owner is changed.

bmahabirbu:
Updated Permissions on both graph rag and pedantic recipe with security measures in mind. I chowned to the default user.

rhatdan (approved):
LGTM

PR #858, opened 2025-03-10
feat: Add Node.js Function calling example using langgraph.js

lholmquist:
Based on our previous email discussions, I've created this node.js function calling recipe using langchain.js and langgraph.js.


recipes/natural_language_processing/function-calling-nodejs/app/Containerfile
```diff
@@ -0,0 +1,65 @@
+# # Install dependencies only when needed
```
jeffmaury: Do we need to keep those commented instructions, they make it hard for a newcomer to read and understand the content

recipes/natural_language_processing/function-calling-nodejs/app/Containerfile
```diff
@@ -0,0 +1,65 @@
...
+
+# EXPOSE 8501
+
+# ENV PORT 8501
+
+# CMD ["sh", "-c", "HOSTNAME=0.0.0.0 node server"]
+
+
+# Install the app dependencies in a full UBI Node docker image
+FROM registry.access.redhat.com/ubi8/nodejs-20:latest
```
jeffmaury: Replace latest by the latest version, dependabot will update it automatically

recipes/natural_language_processing/function-calling-nodejs/app/Containerfile
```diff
@@ -0,0 +1,65 @@
...
+
+# Copy package.json
+COPY package*.json ./
+COPY ./ ./
+
+USER 0
```
jeffmaury, line 8: We should prevent using root

recipes/natural_language_processing/function-calling-nodejs/app/Containerfile
lholmquist: probably not.  I think i was going to remove them and forgot.  let me update

recipes/natural_language_processing/function-calling-nodejs/app/Containerfile
```diff
@@ -0,0 +1,19 @@
+# Install the app dependencies in a full UBI Node docker image
+FROM registry.access.redhat.com/ubi8/nodejs-20
```
jeffmaury:
```suggestion
FROM registry.access.redhat.com/ubi8/nodejs-20:9.5-1741165622
```

jeffmaury (commented):
Also can we use ubi9 instead of ubi8 ?

recipes/natural_language_processing/function-calling-nodejs/app/Containerfile
lholmquist: addressed here https://github.com/containers/ai-lab-recipes/pull/858/commits/f198fdd8a0e0664c5936fe3815b6ac93ecfe38c7

recipes/natural_language_processing/function-calling-nodejs/app/Containerfile
lholmquist: we need to switch to this user to install the dependencies,  but we should be able to change the user after words.  what would be a good user to switch to?

jeffmaury: nodejs image has user default (by default :))

jeffmaury: But the build is ok if you remove this USER instruction

lholmquist: If i remove that `USER 0` then during the npm install there are permission errors with the package-lock.json.  I updated the container file to switch to `USER 1001` after all that happens, similar to what we have in the container file for the other Node.js examples: https://github.com/containers/ai-lab-recipes/blob/main/recipes/natural_language_processing/chatbot-nodejs/app/Containerfile#L50

jeffmaury (approved):
LGTM

rhatdan:
LGTM

PR #892, opened 2026-02-13
fix: added setuptools to req for rag

bmahabirbu:
fixed typo in in rag yaml for port


bmahabirbu:
Need to test on linux, this fixes the error using rag chatbot sayting setup tools is not installed

bmahabirbu:
@jeffmaury PTAL if you have some free cycles

